PR #5, opened 2024-12-09
chore: add examples and refactoring

mahsumdemirwb:


mahsumdemirwb:
Hey guys

I want to take your attention to a couple of things in this PR.

- MotionGroup.stream_run -> the user can give movement_controller as parameter to fully control the websocket communication
- types.py file, Some types from SDK is pretty long to read, so I created a types file which just reassigns some more readable type names


Some ideas I didnt implement but want to share
- I think execute trajectory API is the most important integration point for the end user here, in this API, we can do followings:
   - increase/decrease velocity
   - start/stop movement
   - move forward or backwards
   
basically we open the websocket and then we can send several types of requests and get several types of responses.
Currently I am taking a parameter called `movement_controller` in the stream_run function, in this parameter user can implement whatever they want, so we are exposing full power of the API via SDK and we are not limiting the user.

However, we can make the user's life even easier here, Imagine an API like this:


```
MovementController
.init() # this will send the InitRequest and take the response via websocket
.start(settings) # this will start the movement
.consumeResponse() # this will consume response of movement and decide to add additional logic, e.g. when the robot reached the halway of the trajectory increase the velocity

```

examples/04_move_multiple_robots.py
```diff
@@ -2,6 +2,8 @@
 import asyncio
 import numpy as np
 
+from nova.core.movement_controller import speed_up
```
biering, line 5: Please see `nova` as the public interface. Means we should try to only import from `nova` and not its submodules. If something is not made public yet, we should thing about making it public to the root package.

nova/types/types.py
```diff
@@ -0,0 +1,24 @@
+# This is the stream of requests that is send to execute trajectory websocket
+from typing import AsyncGenerator, Callable
+import wandelbots_api_client as wb
+
+from nova.types.movement_controller_context import MovementControllerContext
+
+ExecuteTrajectoryRequestStream = AsyncGenerator[wb.models.ExecuteTrajectoryRequest, None]
```
biering, line 7: I would put those things into the `__init__.py` file and try to keep manuell mapping as less as possible

nova/types/action.py
```diff
@@ -448,3 +448,18 @@ def orientations(self):
 
     def __add__(self, other: "CombinedActions") -> "CombinedActions":
         return CombinedActions(items=self.items + other.items)
+
+
+    def to_motion_command(self) -> list[wb.models.MotionCommand]:
```
biering, line 453: I actually don't like all those `to_x` methods. Let me think what we can do to reduce mapping. Also the customer should not be required to do any type conversions in order to use the SDK, so we might can make those methods private at least.

nova/types/types.py
mahsumdemirwb: I agree, moving it to the init file sounds like a good idea. It’s a logical place for defining types.

Regarding keeping type mappings minimal:

I think we should avoid unnecessary type mappings, especially if the code is already readable. 
However, we shouldn’t sacrifice code readability to reduce the number of type mappings. Imo, type mapping is a very lightweight abstraction layer, and Python makes re-naming types very easy
 
the wandelbots_api_client has many auto-generated types that are difficult to understand from its name:
 
 - InitializeMovementResponseInitResponse
 - Callable[[AsyncGenerator[wb.models.ExecuteTrajectoryRequest, None]], AsyncGenerator[wb.models.ExecuteTrajectoryResponse, None]] # this is the function type the api_client expects
 
 - wb.models.PlanSuccessfulResponse -> from the name this looks like a response for the planning but this is actually response we get when we load a plan


Another potential solution could be to improve the auto-generation logic to produce more meaningful type names. I think this approach would be more time-consuming compared to Python’s type aliasing features.

nova/types/action.py
mahsumdemirwb: I agree that to_x can grow too much too fast :D 
looking forwards for improvements to the mappings

nova/core/motion_group.py
```diff
@@ -160,121 +151,36 @@ async def _get_trajectory_sample(
...
-        plan_response = await self.plan(actions, tcp)
-
-        # if not rae_pb_parser.plan_response.is_executable(plan_response):
-        #     raise MotionException(plan_response, [], [])
-
-        # await self._get_trajectory_sample(response_rate_in_ms)
-        # TODO: take velocity override into account.
-        # if len(trajectory.sample) > 0 and not math.isnan(trajectory[-1].time):
-        #    self._execution_duration += trajectory[-1].time
+    async def load(self, joint_trajectory: wb.models.JointTrajectory) -> wb.models.PlanSuccessfulResponse:
```
biering, line 155: This method is not really descriptive and it's public. I would rename it to `load_planned_motion` and make it private if possible

mahsumdemirwb: I agree, will implement this

nova/core/motion_group.py
```diff
@@ -140,7 +130,8 @@ async def plan(
             failed_response = plan_response.response.actual_instance
             raise PlanTrajectoryFailed(failed_response)
 
-        return plan_response
+        return plan_response.response.actual_instance
+
 
     async def _get_trajectory_sample(
```
biering, line 135: This method is not used anymore. Can we delete it?

nova/core/motion_group.py
```diff
@@ -37,43 +42,42 @@ async def stream_run(
...
 
-        """
-        if not isinstance(actions, list):
-            actions = [actions]
 
-        # TODO get default tcp if tcp is not set
-        motion_iter = self._planned_motion_iter(
-            actions=actions, tcp=tcp, collision_scene=None, response_rate_in_ms=response_rate_in_ms
+        # EXECUTE MOTION
+        movement_controller_context = MovementControllerContext(
```
biering, line 75: Why do you have that extra context and not passing the parameters in directly?

mahsumdemirwb: My main motivation here is to make sure we have more control and have easy time when we want to add/delete things.

I implemented this by having a Context class which hold all the information that will be passed to the user at this line:
`
_movement_controller = movement_controller(movement_controller_context)
`

the type of movement_controller is this:
`# This is a function which takes a context and returns movement controller
MovementController = Callable[[MovementControllerContext], MovementControllerFunction]
`

so when new fields are added no change is happing in the type hence user doesnt need to change anything.


TBH, I am not sure if this is the best way of doing this in python. Maybe I could have used kwargs? or some other methods?

I am okay with any solution that will fulfill this requirement:
` we should be able to do non-breaking changes on the stream_run function, and we should be able to get a custom movement controller from the user`

nova/core/motion_group.py
mahsumdemirwb: yes I think we can delete this one, will do

examples/02_plan_and_execute.py
```diff
@@ -19,12 +21,34 @@ async def main():
         current_pose = await motion_group.tcp_pose("Flange")
         target_pose = current_pose @ Pose((100, 0, 0, 0, 0, 0))
 
-        actions = [jnt(home_joints), ptp(target_pose), jnt(home_joints)]
+        actions = [
+            jnt(home_joints),
+            ptp(target_pose),
+            jnt(home_joints),
+            ptp(target_pose @ Pose((200, 0, 0, 0, 0, 0))),
```
functionistic, line 28: What about allowing:
`target_pose @ (200, 0, 0, 0, 0, 0)`

mahsumdemirwb: looks great to me, on it

examples/02_plan_and_execute.py
```diff
@@ -2,14 +2,16 @@
...
+
 
 async def main():
     nova = Nova()
     cell = nova.cell()
     controller = await cell.controller("ur")
 
     # Define a home position
-    home_joints = (0, -np.pi / 2, -np.pi / 2, -np.pi / 2, np.pi / 2, 0)
+    home_joints = (0, -np.pi / 4, -np.pi / 4, -np.pi / 4, np.pi / 4, 0)
```
functionistic, line 14: `np.pi` can we make that nicer?

examples/03_move_and_set_ios.py
```diff
@@ -34,10 +34,10 @@ async def main():
         # plan_response = await motion_group.plan(trajectory, tcp="Flange")
         # print(plan_response)
 
-        motion_iter = motion_group.stream_run(actions, tcp="Flange")
-        async for motion_state in motion_iter:
-            print(motion_state)
+        def print_motion(motion):
+            print(motion)
 
+        await motion_group.stream_run(actions, tcp="Flange", initial_movement_consumer=print_motion)
```
biering, line 40: doesn't return this a stream that needs to be iterated? shouldn't it called be run?

examples/02_plan_and_execute.py
mahsumdemirwb: sure, should we expose a pi from nova?
e.g. 
`
from nova import pi
`
and home position definition stays same

technically, they need to add some float values there so, I can also make like this:
home_joints = (0, -1, -1.123123, ...)

so no usage of pi

examples/04_move_multiple_robots.py
```diff
@@ -22,7 +22,9 @@ async def move_robot(controller: Controller):
         target_pose = current_pose @ Pose((100, 0, 0, 0, 0, 0))
         actions = [jnt(home_joints), ptp(target_pose), jnt(home_joints)]
 
-        await motion_group.run(actions, tcp="Flange")
+        await motion_group.stream_run(
```
biering, line 25: same here

examples/03_move_and_set_ios.py
mahsumdemirwb: hmm, good idea, updating the function name to just `run`

examples/02_plan_and_execute.py
mahsumdemirwb: done, also updated the second example where I use a list, a tuple and a function which yields numbers

examples/02_plan_and_execute.py
```diff
@@ -2,29 +2,52 @@
...
+        ]
 
         # plan_response = await motion_group.plan(trajectory, tcp="Flange")
         # print(plan_response)
 
-        await motion_group.run(actions, tcp="Flange")
+        await motion_group.run(actions, tcp="Flange", movement_controller=move_forward)
+
+
+def as_iterator(data: list | tuple):
```
biering, line 48: I'm not sure about the usefulness in this example. I would remove this one and prefer to have that in an unit test.

examples/04_move_multiple_robots.py
```diff
@@ -13,7 +13,7 @@ async def main():
 
 
 async def move_robot(controller: Controller):
```
biering, line 15: Just for readability I would like to switch move_robot and main method

biering: I mean the code positions

mahsumdemirwb: sure, done :D

examples/02_plan_and_execute.py
mahsumdemirwb: I think unit test is better to cover all the possible cases, for now I removed this but didnt add any unit tests, 
I covered this case also in the doctests

PR #9, opened 2024-12-11
fix: not always use "Flange" as a parameter for tcp in _load_planned_motion

rksr:
Fix the `_load_planned_motion` method to accept a `tcp` and not always use `Flange` as a parameter for tcp. Adjust the call to this method in the `run` function accordingly.


biering:
Looks good! I have to adjust the PR title check a bit

PR #11, opened 2024-12-17
feat(RPS-999): activate motion groups seperatly

mahsumdemirwb:


nova/core/motion_group.py
```diff
@@ -18,12 +18,35 @@
...
+    ):
         self._api_gateway = api_gateway
         self._motion_api_client = api_gateway.motion_api
         self._cell = cell
         self._motion_group_id = motion_group_id
         self._current_motion: str | None = None
+        self.is_activated = is_activated
+
+    async def __aenter__(self):
+        if self.is_activated:
```
biering, line 32: Is it possible to read the state from the motion group? I would not want to risk that this gets out of sync

examples/05_selection_motion_group_activation.py
```diff
@@ -0,0 +1,43 @@
...
+async def main():
+    nova = Nova()
+    cell = nova.cell()
+    ur = await cell.controller("ur")
+    kuka = await cell.controller("kuka")
+
+    ur_0_mg = ur.get_motion_group()
+    kuka_0_mg = kuka.get_motion_group()
+
+    async with AsyncExitStack() as stack:
```
biering, line 35:
```suggestion
    async with (ur.motion_group() as ur_0_mg, kuka.motion_group() as kuka_0_mg):
```

nova/core/controller.py
```diff
@@ -26,7 +26,10 @@ async def __aenter__(self):
         for mg in motion_groups:
```
biering, line 26: Proposal: Create a private method that fetches all motion groups from that controller.

```
def _fetch_motion_groups(reload = False):
  if reload or self._motion_groups is None:
    fetch motion groups from API
  return self._motion_groups
```

.motion_group()
```
def motion_group(...):
    motion_groups = self._fetch_motion_groups()
    return self._motion_groups[key]
```

biering: if we can really make the .motion_group sync we should implement the index access again
```python
mg_0 = controller[0]
```

and
```
with controller[0] as mg_0:
    ...
```

PR #15, opened 2024-12-20
feat(RPS-1004): handle https, add some convenience methods

biering:
- handle https and improve gateway creation
- conform env variables from ADX
- added `tcps()` and `tcp_names()` method
- added `controllers()` method
- make tcp in `get_state()` optional -> returns flange pose
- updated examples accordingly


mahsumdemirwb (approved):

langenhagen (approved):
lgtm

examples/03_move_and_set_ios.py
```diff
@@ -10,17 +10,18 @@
...
+    controllers = await cell.controllers()
 
     # Define a home position
     home_joints = (0, -pi / 4, -pi / 4, -pi / 4, pi / 4, 0)
 
     # Connect to the controller and activate motion groups
-    async with controller:
-        motion_group = controller.motion_group()
+    async with controllers[0] as motion_group:
+        tcp_names = await motion_group.tcp_names()
```
bompo, line 20: I guess TCP is name only. But would also be nice if its straight an object with the pose offset

examples/02_plan_and_execute.py
```diff
@@ -10,15 +10,19 @@
...
-    controller = await cell.controller("ur")
+    controllers = await cell.controllers()
+    controller = controllers[0]
 
     # Define a home position
     home_joints = (0, -pi / 4, -pi / 4, -pi / 4, pi / 4, 0)
 
     # Connect to the controller and activate motion groups
-    async with controller[0] as mg:
+    async with controller[0] as motion_group:
```
bompo, line 20: `with controller as motion_group`
Thinking about it, it's a bit strange that contoller is an array itself. Maybe a controller.defaultMotionGroup is more explicit

nova/core/motion_group.py
```diff
@@ -44,6 +44,32 @@ def current_motion(self) -> str:
         #    raise ValueError("No MotionId attached. There is no planned motion available.")
         return self._current_motion
 
+    async def plan(self, actions: list[Action], tcp: str) -> wb.models.JointTrajectory:
+        current_joints = await self.joints(tcp=tcp)
+        robot_setup = await self._get_optimizer_setup(tcp=tcp)
```
bompo, line 49: could be cached?

nova/core/motion_group.py
```diff
@@ -176,3 +162,36 @@ async def stop(self):
...
+        Args:
+            tcp (str): The identifier of the tool center point (TCP) to be used for tcp_pose in response. If not set,
+                the flange pose is returned as tcp_pose.
+        """
+        response = await self._api_gateway.motion_group_infos_api.get_current_motion_group_state(
+            cell=self._cell, motion_group=self.motion_group_id, tcp=tcp
+        )
+        return response
+
+    async def joints(self, tcp: str | None = None) -> wb.models.Joints:
```
bompo, line 178: IDK why TCP is offered as a parameter at all? It doesn't make a difference for the joint positions? I would remove it.

nova/core/motion_group.py
```diff
@@ -176,3 +162,36 @@ async def stop(self):
             logger.debug(f"Motion {self.current_motion} stopped.")
         except ValueError as e:
             logger.debug(f"No motion to stop for {self}: {e}")
+
+    async def get_state(self, tcp: str | None) -> wb.models.MotionGroupStateResponse:
```
bompo, line 166: Is there also the option for a state fetching via websocket?

nova/gateway.py
```diff
@@ -76,15 +78,20 @@ def __init__(
         if access_token is None:
             access_token = config("NOVA_ACCESS_TOKEN", default=None)
 
+        if (username is None or password is None) and access_token is None:
```
bompo, line 81: What happens if all three are set? Access_token should have prio.

bompo (approved):

examples/02_plan_and_execute.py
biering: The index access is just convenience. You can also do:
```python
with controller.motion_group(0) as motion_group:
    ...
```

nova/core/motion_group.py
biering: Yeah true, for joints it should not make a difference.

examples/03_move_and_set_ios.py
biering: yes, but for that I've added the `.tcps()` method which returns a list of tcp objects

nova/gateway.py
biering: fixed

nova/core/motion_group.py
biering: done

nova/core/motion_group.py
biering: not yet (besides the auto-generated API), but we should tackle that in a new MR.

PR #17, opened 2025-01-06
feat: added mypy check to CI * updated API * added pose transform

biering:


nova/actions.py
```diff
@@ -20,8 +20,8 @@ class WriteAction(Action):
     value: Any
 
     @pydantic.model_serializer
-    def serialize_model(self) -> wb.models.IOValue:
-        return wb.models.IOValue(io=self.key, boolean_value=self.value)
+    def serialize_model(self):
```
functionistic, line 23: Why no return type anymore?

biering: Because this method always needs to return a dict (because of pydantic). It was wrong before.

PR #22, opened 2025-01-10
feat: extend Vector3d

langenhagen:
Give it additional capabilities that formerly was split between `Orientation` and
`Position` types. We decided to consolidate those into `Vector3d` because,
technically, all of them are just 3-vectors. The semantic
differentiation may be nice on paper but turned out to be unnecessarily
complex and all-over-the-place. Instead, have one powerful vector3d type
for all of them.

The wandelscript repo directly benefits from that due to increased simplicity.

Consequently, add geometricalgebra as a dependency. I copied/frankensteined the code basically out of the wandelbrain.

Also allow for some new, common functions like negation and substractions.
I believe that's a good call for a public class in an SDK.

Also add `darker` as a dependency package because its just nice to black-format your changeset before committing


mahsumdemirwb (approved):

pyproject.toml
```diff
@@ -32,6 +32,7 @@ pytest = "^8.3.4"
 pytest-asyncio = "^0.24.0"
 mypy = "^1.13.0"
 pre-commit = "^4.0.1"
+darker = {extras = ["isort"], version = "^2.1.1"}
```
biering, line 35: You can sort imports with ruff via
```bash
ruff check --select I --fix
```

I would use ruff and don't add another dev dependency

langenhagen: cool beans, added a commit that incorporates that into `.pre-commit.yaml` so this gets run before each commit. Feels good, speedwise.

PR #23, opened 2025-01-13
feat: add support for motion setting

mahsumdemirwb:


examples/08_move_with_settings_planner.py
```diff
@@ -0,0 +1,56 @@
...
+        tcp_names = await motion_group.tcp_names()
+        tcp = tcp_names[0]
+
+        current_pose = await motion_group.tcp_pose(tcp)
+
+        # Start building a trajectory
+        trajectory = Trajectory()
+
+        # You can provide settings which are applied to that specific motion only
+        trajectory.move(via="jnt", to=home_joints, velocity=10)
```
biering, line 30: I would do this a bit differnent. `move` implies that the robot already moves at that point. But actually it builds the planner trajectory. 

So I would rather propose something like

```python
planner.jnt(target, settings=(...))
```

langenhagen (approved):
apart from what @biering  said, looks good to me

examples/08_move_with_settings_planner.py
langenhagen: Should I understand this as an issue in our SDK's API?

nova/motion_settings.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # max_tcp_orientation_velocity in the API
+    orientation_velocity: float | None = pydantic.Field(default=None)
+
+    # max_tcp_orientation_acceleration in the API
+    orientation_acceleration: float | None = pydantic.Field(default=None)
+
+    @classmethod
+    def field_to_varname(cls, field):
+        return f"__ms_{field}"
```
functionistic, line 38: AFAIK this is WS related only and should be moved there.

nova/motion_settings.py
```diff
@@ -0,0 +1,85 @@
...
+            tcp_velocity_limit=self.velocity,
+            tcp_acceleration_limit=self.acceleration,
+            tcp_orientation_velocity_limit=self.orientation_velocity,
+            tcp_orientation_acceleration_limit=self.orientation_acceleration,
+        )
+
+    def as_blending_setting(self) -> wb.models.MotionCommandBlending:
+        # don't allow setting both, API only accepts one at a time
+        if self.min_blending_velocity and self.blending:
+            raise ValueError("Can't set both min_blending_velocity and blending")
```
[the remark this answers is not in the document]
functionistic: Also these `as_` methods introduce a dependency to `wb.models` which I don't like in a plain data object. Not a big issue atm but should maybe considered.

nova/motion_settings.py
```diff
@@ -0,0 +1,85 @@
+import pydantic
+import wandelbots_api_client as wb
+
+
+class MotionSettings(pydantic.BaseModel):
+
```
functionistic, line 6: Missing the docblock from before the line based comments below might be okay but these will not be recognised by documentation generators.

nova/motion_settings.py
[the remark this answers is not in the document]
mahsumdemirwb: First point: validation/hoos:

I aggree, implemented

Second points: mapping

I aggree that keeping data classes lean is good. I would like to figure out a solution which does this. I can't come up with a quick fix to this because, the class which is using MotionSetting is also a pydantic class, so If I can't move the logic there also and this logic needs to live somewhere :D

I looked into the dataclasses.asdict also but I couldn't figure out a solution that gives us what we want.
since you said it is not a big issue ATM I would like leave this to offline discussion

nova/motion_settings.py
mahsumdemirwb: I also couldn't see a usage of this in the SDK, so removed

nova/motion_settings.py
mahsumdemirwb: updated the documentaion, about the field names, here is the trade-offs I see:
- shorter is easier to use
- longer is more expressive
- longer is closer to the API
- shorter is closer to wandelscript

I think we need to decide which trade-offs are more important for us.

@biering @langenhagen can you also provide what is your preference?

biering: Both options are valid to me but when we need a decision I would go with the names from the API to make them verbose. I think the people from robotics thought a lot about the names and made a good decision. 

We are still free to don't use the exact naming in Wandelscript if it makes sense to us to have it more concise or to adapt Wandelscript to it.

langenhagen: I second the idea to go with the names from the API to make them verbose. Clients/Users can always alias names away and maybe we come up with concise aliases later on.

As guiding wiscecrack-observation: Code is many times more often read than written.

examples/07_move_with_settings_plain.py
```diff
@@ -0,0 +1,56 @@
...
+            ptp(target_pose @ [100, 0, 0, 0, 0, 0]),
+            jnt(home_joints),
+            ptp(target_pose @ (100, 100, 0, 0, 0, 0)),
+            jnt(home_joints),
+            ptp(target_pose @ Pose((0, 100, 0, 0, 0, 0))),
+            jnt(home_joints),
+        ]
+
+        # you can update the settings of the action
+        for action in actions:
```
biering, line 48: Since the examples are basically the first entry point for the users we need to keep the number as low as possible to not overwhelm the user. I would put this settings modification in the 02 example, or what do you think?

mahsumdemirwb: I agree, will refactor

nova/motion_settings.py
mahsumdemirwb: thanks for the inputs, will update to use the longer names

functionistic (approved):

PR #25, opened 2025-01-14
feat(RPS-976): added integration test CI

biering:


nova/core/nova.py
```diff
@@ -47,24 +49,135 @@ def __init__(self, api_gateway: ApiGateway, cell_id: str):
...
+        Args:
+            name (str): The name of the controller
+
+        Return: models.ControllerInstance: The controller instance
+        """
+        controllers = await self._get_controller_instances()
+        controller = next((c for c in controllers if c.controller == name), None)
+        return controller
+
+    async def _wait_for_controller_to_be_ready(self, name: str, timeout: int):
```
functionistic, line 68: Could also be:
```suggestion
    async def _wait_for_controller_ready(self, name: str, timeout: int):
```

functionistic (approved):

PR #32, opened 2025-01-22
feat: expose error object in exception

stefanwagnerdev:


biering (approved):

nova/core/exceptions.py
```diff
@@ -10,14 +10,32 @@ def __init__(self, controller: str):
 
 class PlanTrajectoryFailed(Exception):
     def __init__(self, error: wb.models.PlanTrajectoryFailedResponse):
+        self._error = error
         super().__init__(f"Plan trajectory failed: {json.dumps(error.to_dict(), indent=2)}")
```
biering, line 14: Isn't that `error.model_dump()`?

stefanwagnerdev: I don't think there is a need for serialization. Just pass the object should be fine

biering: Ups, I mean in the next line

stefanwagnerdev: No I think this is fine, at least it shows me the error nicely:

stefanwagnerdev: Bildschirmfoto 2025-01-22 um 14 19 50

PR #34, opened 2025-01-23
feat: Make authorization flow asynchron.

sn3ek:
This merge request updates the authorization process to be asynchronous instead of sync busy loop.

### DIFF
--- a/examples/07_auth0_with_device_code.py
+++ b/examples/07_auth0_with_device_code.py
@@ -1,4 +1,5 @@
 from nova.auth.authorization import Auth0DeviceAuthorization
+import asyncio
 
 """
 Example: Perform device authorization with Auth0.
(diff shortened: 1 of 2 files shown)


examples/07_auth0_with_device_code.py
```diff
@@ -8,14 +9,14 @@
 """
 
 
-def main():
+async def main():
     # Replace these values with your Auth0 domain, client ID, and audience
-    auth0_domain = "YOUR-AUTH0-DOMAIN"
-    auth0_client_id = "YOUR-AUTH0-CLIENT-ID"
-    auth0_audience = "YOUR-AUTH0-AUDIENCE"
+    AUTH0_DOMAIN = "auth.portal.stg.wandelbots.io"
```
langenhagen, line 14: sir, this is a public repo. Is this deliberate?

langenhagen (commented):
Looks sound, except for the authentication details in our public repo

langenhagen (changes requested):
s.a.

examples/07_auth0_with_device_code.py
sn3ek: We have to had those information out to the customers. But i removed them and we will put them as variables into the github pipleine to generate them as env variables.

CLIENT_SECRET would be problematic

poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
langenhagen, line 1: nit: the poetry.lock file could have stayed as is,

langenhagen (approved):
lgtm. Curious that the lockfile changed, seem benign though.

poetry.lock
sn3ek: Reverted it - this was because of asyncio dep. But i reverted it as well.

mahsumdemirwb:
Tested, works for me

nova/auth/authorization.py
```diff
@@ -156,34 +166,32 @@ def poll_token_endpoint(self):
...
-                    self.interval += 5
-                    time.sleep(self.interval)
-            elif token_response.status_code == 403:
-                # If the response status is 403, it means the request is forbidden, wait and retry.
-                print("Waiting...")
-                time.sleep(self.interval)
-            else:
-                # For other status codes, raise an exception with the error details.
-                raise Exception("Error:", token_response.status_code, token_response.json())
+        async with httpx.AsyncClient() as client:
```
mahsumdemirwb, line 169: I think the amount of nesting here is a sign of refactoring, but since it is SDK-internal code, this can be done later, and the code is fairly simple to refactor later

mahsumdemirwb (approved):

PR #42, opened 2025-02-06
fix: simple check for http prefix

andiikaa:
https://wandelbots.atlassian.net/browse/RPS-1208

### DIFF
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -1,4 +1,5 @@
 import os
+
 import pytest
 
 from nova.gateway import ApiGateway


mahsumdemirwb (approved):

nova/gateway.py
```diff
@@ -127,3 +128,8 @@ def __init__(
 
     async def close(self):
         return await self._api_client.close()
+    
+    def _host_with_prefix(host: str) -> str:
+        if host.startswith("http"):
+            return host
+        return f"http://{host}"
```
biering, line 135: In this case when NOVA_API is something like `https://...` it will become `http://https://...`. Is this intended?

andiikaa: But the string `https://...` is starting with `http` thus the host will be returned directly, without prefix. Or did i overlook something?

biering: Ahh true, yes

nova/gateway.py
```diff
@@ -127,3 +128,8 @@ def __init__(
 
     async def close(self):
         return await self._api_client.close()
+    
+    def _host_with_prefix(host: str) -> str:
```
biering, line 132:
```suggestion
    def _host_with_prefix(host: str) -> str:
       """
       Examples:
       >>> _host_with_prefix("https://example.com")
       https://example.com
       >>> _host_with_prefix("example.com")
       http://example.com
       """
```

maybe add some doctest to the function like this above

PR #43, opened 2025-02-06
feat: streaming execution first draft

functionistic:
This is the first draft to make `execute()` and `plan_and_execute()` AsyncGenerators i.e. streaming.
It is intended as a quick and somewhat dirty solution to enable porting wandelscript execution as is.

### DIFF
--- a/nova/core/movement_controller.py
+++ b/nova/core/movement_controller.py
@@ -1,3 +1,5 @@
+from typing import Callable
+
 import wandelbots_api_client as wb
 from loguru import logger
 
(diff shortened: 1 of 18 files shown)


nova/core/robot_cell.py
```diff
@@ -202,6 +202,24 @@ def state_stream(self, rate: int) -> AsyncIterable[AbstractDeviceState]:
         """
 
 
+class Callerator():
```
biering, line 205: The name is great :D

nova/core/movement_controller.py
```diff
@@ -42,7 +42,7 @@ def motion_group_state_to_motion_state(
 
 
 def move_forward(
-    context: MovementControllerContext, on_movement: Callable[[MotionState], None] | None
+    context: MovementControllerContext, on_movement: Callable[[MotionState | None], None] | None
```
biering, line 45: is this still needed?

functionistic: For the current streaming approach, yes.

PR #64, opened 2025-02-11
feat(RPS-1221): refine nova actions interface

biering:


stefanwagnerdev:
This touches quite a lot of code which is in https://github.com/wandelbotsgmbh/wandelbots-nova/pull/61

biering:
@stefanwagnerdev I've synced with Mahsum on this. It will be quickly fixed but it's important to make that consistent before the first customer is using it. I also updated the bridge

PR #66, opened 2025-02-12
feat(RPS-1224): added user-agent to api client

biering:


biering:
I've tested it manually against an instance

PR #68, opened 2025-02-12
feat(RPS-1214): streaming execution

functionistic:


nova/core/robot_cell.py
```diff
@@ -323,38 +322,69 @@ async def execute(
...
-                movement_controller=movement_controller,
-                on_movement=callerator,
-            )
+        def unpack_movement_response(
+            movement_response: MovementResponse,
+        ) -> Any:  # TODO: can we use a more specific type here?
+            if isinstance(movement_response, nova_models.ExecuteTrajectoryResponse):
+                return movement_response.actual_instance
+            # TODO: handle the StreamMoveResponse case or make sure it doesn't happen
+            assert False, f"Unexpected movement response: {movement_response}"
```
langenhagen, line 331: nit: some linters complain about the use of `assert` in production code, whereas `raise AssertionError()` goes fine with the linters.

langenhagen (approved):
lgtm

nova/core/motion_group.py
```diff
@@ -354,22 +355,32 @@ async def _execute(
...
+                and instance.standstill.reason == wb.models.StandstillReason.REASON_MOTION_ENDED
+            )
+
+        execute_response_streaming_controller = StreamExtractor(controller, stop_condition)
+        execution_task = asyncio.create_task(
+            self._api_gateway.motion_api.execute_trajectory(
+                self._cell, execute_response_streaming_controller
+            )
+        )
+        async for execute_resposne in execute_response_streaming_controller:
```
biering, line 381:
```suggestion
        async for execute_response in execute_response_streaming_controller:
```

nova/core/robot_cell.py
```diff
@@ -21,17 +20,19 @@
     runtime_checkable,
 )
 
-import aiostream
 import anyio
 import asyncstdlib
 import pydantic
+import wandelbots_api_client as wb
```
biering, line 26: since you already use nova.api

I would only use `from nova import api` and use that for the rest

biering (approved):

nova/core/robot_cell.py
functionistic: It's not meant to stay there. I will test for removal after #76 is merged

PR #76, opened 2025-02-14
fix: collision-free-movement-type

mahsumdemirwb:


nova/core/motion_group.py
```diff
@@ -266,62 +286,48 @@ async def _plan_collision_free(
...
+                    actions=batch,
+                    tcp=tcp,
+                    start_joint_position=current_joints,
+                    optimizer_setup=robot_setup,
                 )
                 all_trajectories.append(trajectory)
+                # the last joint position of this trajectory is the starting point for the next one
                 current_joints = tuple(trajectory.joint_positions[-1].joints)
 
         # TODO: combine all trajectories into one
```
biering, line 327: isn't this done?

mahsumdemirwb: yes, it is, removed

examples/08_multi_step_movement_with_collision_free.py
```diff
@@ -0,0 +1,65 @@
...
+                ptp(target_pose @ Pose((0, 50, 0, 0, 0, 0))),
+                jnt(home_joints),
+            ]
+
+        # you can update the settings of the action
+        for action in actions:
+            action.settings = MotionSettings(tcp_velocity_limit=200)
+
+        joint_trajectory = await motion_group.plan(
+            actions, tcp, start_joint_position=(-0.0429, -1.8781, 1.8464, -2.1366, -1.4861, 1.0996)
```
functionistic, line 56: Where are these joint positions from? The rest of the program depends on the initial position of the virtual robot.

examples/08_multi_step_movement_with_collision_free.py
```diff
@@ -0,0 +1,65 @@
...
+            tcp_names = await motion_group.tcp_names()
+            tcp = tcp_names[0]
+
+            # Get current TCP pose and offset it slightly along the x-axis
+            current_pose = await motion_group.tcp_pose(tcp)
+            target_pose = current_pose @ Pose((100, 0, 0, 0, 0, 0))
+
+            actions = [
+                ptp(target_pose),
+                collision_free(home_joints),
```
stefanwagnerdev, line 42: How does collision_free movement work without a defined collision scene

nova/core/movement_controller.py
```diff
@@ -69,10 +69,8 @@ async def movement_controller(
                 raise InitMovementFailed(r1.init_response)
 
         # The second request is to start the movement
-        set_io_list = context.combined_actions.to_set_io()
-        yield wb.models.StartMovementRequest(
-            set_ios=set_io_list, start_on_io=None, pause_on_io=None
-        )  # type: ignore
+        # set_io_list = context.combined_actions.to_set_io()
```
functionistic, line 72: How are IOs set now?

examples/08_multi_step_movement_with_collision_free.py
mahsumdemirwb: This is a joint position I took from the robot pad. It should be different from the first position in the trajectory; apart from that, it is an arbitrary point

> The rest of the program depends on the initial position of the virtual robot.
If we don’t provide the home position, it will still work, but it will not trigger move_to_start_position.

examples/08_multi_step_movement_with_collision_free.py
mahsumdemirwb: Collision scene fields are optional in the "Plan Collision Free PTP" API. This is why we made the collision scene parameters optional

nova/core/movement_controller.py
mahsumdemirwb: This one is missing :( and will take some work to add back, but I am on it and will ask review again after implementing this

examples/08_multi_step_movement_with_collision_free.py
stefanwagnerdev: Ok, but the call is quite useless in that case as it also could be a simple p2p in that case.

examples/08_multi_step_movement_with_collision_free.py
stefanwagnerdev: I would prefer if the examples only "plan" and the execution is optional. That way they would be independant from the current robot config and always "work".

examples/08_multi_step_movement_with_collision_free.py
mahsumdemirwb: Yes It can be p2p but then it wouldn't be a mixed action list which is what I am trying to have

examples/08_multi_step_movement_with_collision_free.py
mahsumdemirwb: I agree that executing trajectory can bring additional code flows which can fail but this would be good for us becauase we are using examples in ci/cd tests, we will see the error and fix it

In this specific example, If I remove execution I would loose below things:
- move to start postion not tested
- combined trajectory execution is not tested

nova/core/movement_controller.py
mahsumdemirwb: fixed now, tested with simple action lists and complex action lists and saw the IO is both set and also set at correct place

PR #86, opened 2025-02-18
feat: collision free benchmark

stefanwagnerdev:
The benchmark suite evaluates different motion planning strategies against a standardized set of problems from the [robometrics](https://github.com/fishbotics/robometrics) dataset.

Installable as an extra with 
poetry install --extras "benchmark"

Individual benchmarks can be run with:
python nova_rerun_bridge/benchmark/run_collision_free_benchmark.py

The result of the benchmark is logged to the console and the individual recordings are stored in the folder benchmark_results

Screenshot 2025-02-18 at 11 57 32
Screenshot 2025-02-18 at 11 57 24
Screenshot 2025-02-18 at 11 57 15
Screenshot 2025-02-18 at 11 57 05


nova_rerun_bridge/benchmark/log_successful_planning.py
```diff
@@ -0,0 +1,123 @@
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
```
langenhagen, line 3: s.a.

nova_rerun_bridge/benchmark/log_successful_planning.py
```diff
@@ -0,0 +1,123 @@
...
+        "z_axis": (0.125, 0.502, 1.0, 1.0),  # #2080ff - Blue
+        "target": (0, 255, 0, 255),  # Green
+    }
+
+
+@dataclass
+class VisualizationConfig:
+    """Configuration for trajectory visualization."""
+
+    arrow_length: float = 50.0  # mm
```
langenhagen, line 28: :+1:  appreciate the unit comment

nova_rerun_bridge/benchmark/robometrics_helper.py
```diff
@@ -0,0 +1,114 @@
+from typing import Any, Dict, List
```
langenhagen, line 1: s.a.

nova_rerun_bridge/benchmark/run_collision_free_magma.py
```diff
@@ -0,0 +1,144 @@
...
+                                z=target.orientation[2],
+                            ),
+                            tcp=tcp,
+                        ),
+                    ),
+                )
+            )
+            return response.joint_positions
+        except Exception as e:
+            logger.error(f"Failed to calculate inverse kinematics: {str(e)}")
```
langenhagen, line 73: i'm a big fan of logging exceptions as with `logger.exception` because of stacktrace and info about the exception. maybe overkill here. would also eradicate the need for getting a handle to the exception with `... as e:`. just wanted to bring that up. also below.

pyproject.toml
```diff
@@ -27,7 +27,6 @@ wandelbots_api_client = "25.1.0"
 websockets = "^14.1.0"
 loguru = "^0.7.2"
 pydantic = "^2.10.3"
-numpy = "^2.1.3"
```
langenhagen, line 30: confused: you remove numpy from the pyproject.toml, but above you import numpy regularly. I'd be explicit about direct dependencies.

langenhagen (commented):
largely ok, much nice code, very readable. bit big the pr, some nits. the numpy dependency removal irritates me

pyproject.toml
stefanwagnerdev: this forces numpy 2 but the benchmark lib I use wants numpy 1
Our code does not really need to force the upgrade

stefanwagnerdev: I removed the dependency which forced the downgrade and copied the datasets into our project

langenhagen: not what I expected but lgtm

langenhagen (approved):
lgtm but one thing - unless super obvious - would leave the resolution of review threads to the author of the thread or alternatively just leave a comment, e.g. what course of action you went. Now, that you marked the threads as resolved and did not comment on the course of action I had to go back to the code myself and check what you actually did or did not do. Not a biggie but not cool.

Upon a glance type hints are still modernizable - e.g. searching for `from typing` shows altbacken uses of `Optional` and `List` but as Mediocrates said: "Ehh, good enough"

Have my approval.

stefanwagnerdev:
@langenhagen Thanks for pointing this out. 
Ruff can do this as well with
ruff check --select UP006 --fix .
So I added this rule to our linter.

biering (dismissed):
lgtm

biering (dismissed):

PR #91, opened 2025-02-20
chore: simplified splitting

functionistic:
If we think the intermediary list is a problem there is also this generator:
```python
def split_actions_into_batches(actions: list[Action]) -> Generator[list[Action], None, None]:
    current_batch: list[Action] = []
    for action in actions:
        if isinstance(action, CollisionFreeMotion):
            if current_batch:
                yield current_batch
                current_batch = []
            yield [action]
        else:
            current_batch.append(action)
    if current_batch:
        yield current_batch
```
vs
```python
def split_actions_into_batches(actions: list[Action]) -> list[list[Action]]:
    batches: list[list[Action]] = []
    for action in actions:
        if (
            # Start a new batch if:
            not batches  # first action no batches yet
            or isinstance(action, CollisionFreeMotion)
            or isinstance(batches[-1][-1], CollisionFreeMotion)
        ):
            batches.append([action])
        else:
            batches[-1].append(action)
    return batches
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,10 @@ async def main():
       access_token="your-access-token"
   )
   bridge = NovaRerunBridge(nova)
-  
+
   # Setup visualization
   await bridge.setup_blueprint()
-  
+
   # Setup robot
   cell = nova.cell()
   controller = await cell.ensure_virtual_robot_controller(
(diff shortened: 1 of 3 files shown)


mahsumdemirwb (dismissed):

mahsumdemirwb:
This really needed some simplification, thanks!

biering (approved):

PR #92, opened 2025-02-21
feat(docs): Add autodocs to pipeline

NaveedBhuiyan:


langenhagen (commented):
thanks Naveed - 2 points:

1. is this tested? I would see a compiled version of these docs, e.g. by setting the branch to your feature branch once for testing, see if this works, then if yes fixup your commit to react on `main`

2. wasn't aware of us using Github pages for now. Wondering if this is aligned with people - haven't heard of this. 

Thus, I'm generally OK with this change but would wait for your answer before approving

NaveedBhuiyan:
Hey @langenhagen 

Thanks for the quick response. Regarding both your points we would need to enable github pages for testing and execution. So far this has only been tested on a forked repo.

@stefanwagnerdev How should we proceed from here ? Would it be possible to enable github pages ?

stefanwagnerdev:
I think its totally fine to push the api docs to github pages. We will add a high level overview in the developer portal at some point as well.

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: pip install poetry
+
+      - name: Add pdoc to dependencies
+        run: poetry add --dev pdoc
+
+      - name: Install dependencies
+        run: poetry install
+
+      - name: Generate documentation
+        run: poetry run pdoc nova -o ./docs
```
stefanwagnerdev, line 39: maybe also add nova-rerun-bridge here so the bridge functions are also logged

stefanwagnerdev (approved):

.github/workflows/autodocs.yml
NaveedBhuiyan: yup done.

biering (approved):

mahsumdemirwb (approved):

PR #97, opened 2025-02-24
feat: stream robot state to rerun on standard timeline

stefanwagnerdev:


martinfg (commented):
Looks fine overall! Im not too inclined though to use this in that context. I think for a application i would consider logging motions (planning ahead) to a completly different `.rdd` or application instead of logging to a different timeline. But maybe thats just a different use case. I'll have to try both!

stefanwagnerdev:
> Looks fine overall! Im not too inclined though to use this in that context. I think for a application i would consider logging motions (planning ahead) to a completly different `.rdd` or application instead of logging to a different timeline. But maybe thats just a different use case. I'll have to try both!

I guess you could stream this in a new recording file. The latest timeline is shown by default so not much of a difference.

nova_rerun_bridge/examples/08_stream_robot.py
```diff
@@ -2,10 +2,18 @@
...
 from contextlib import asynccontextmanager
 
+from nova import MotionSettings
+from nova.actions import jnt, ptp
 from nova.api import models
 from nova.core.nova import Nova
+from nova.types import Pose
 from nova_rerun_bridge import NovaRerunBridge
 
+"""
```
langenhagen, line 12: nit: docstrings like this usually go to the top of the file, above the imports.

Then, `ruff` might be able to complain about the first line not being on the opening quotion marks and other things. Just a nitpick, though :)

langenhagen (approved):
lgtm

nova_rerun_bridge/examples/08_stream_robot.py
stefanwagnerdev: I moved it up

PR #100, opened 2025-02-24
fix: remove gltf-transform dependency and update model download

stefanwagnerdev:
…process

The models are now attached to the react components release as an extra asset. No need to decompress them anymore.
@NaveedBhuiyan

### DIFF
--- a/.github/workflows/push_nova_rerun_bridge_image.yaml
+++ b/.github/workflows/push_nova_rerun_bridge_image.yaml
@@ -18,12 +18,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


NaveedBhuiyan:
> …process
> 
> The models are now attached to the react components release as an extra asset. No need to decompress them anymore. @NaveedBhuiyan

Awesome
Thank you soo much

NaveedBhuiyan (commented):
Reviewed the code and Tested this locally. "poetry run download-models" runs smoothly now and I can run the examples as well without any additional cmd from gltf transform. Not sure if I have approval rights but since I tested this and it works I will approve this.

NaveedBhuiyan (approved):

PR #104, opened 2025-02-25
feat: implement auto refresh of access token

stefanwagnerdev:
Uses the auth helper to auto refresh the access token if the login is not valid

### DIFF
--- a/nova/core/gateway.py
+++ b/nova/core/gateway.py
@@ -4,6 +4,7 @@ import asyncio
 import functools
 import time
 from abc import ABC
+from dataclasses import dataclass
 from typing import TypeVar
 
 import wandelbots_api_client as wb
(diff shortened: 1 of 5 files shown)


nova/core/gateway.py
```diff
@@ -79,6 +82,9 @@ def __init__(
         access_token: str | None = None,
         version: str = "v1",
         verify_ssl: bool = True,
+        auth0_domain: str | None = None,
+        auth0_client_id: str | None = None,
+        auth0_audience: str | None = None,
```
functionistic, line 87: If all of them are mandatory to create the `Auth0DeviceAuthorization` object later, whynot pass them as a dataclass and simplify the individual None handling below?

stefanwagnerdev: I would rather just add them as default variables?
Are there still concerns about that @sn3ek ?

biering (approved):
pretty cool!

nova/core/gateway.py
stefanwagnerdev: I added the dataclass

langenhagen (approved):
looking good!

nova/core/gateway.py
functionistic: Nice!

PR #107, opened 2025-02-26
chore: Downgrade numpy to 1.x.x

langenhagen:
Upon customer request, downgrade numpy to 1.x.x.

As of today, resolves to numpy==1.26.4 on my machine.


pyproject.toml
```diff
@@ -32,12 +32,12 @@ scipy = "^1.14.1"
 aiostream = "^0.6.4"
 anyio = "^4.8.0"
 asyncstdlib = "^3.13.0"
-numpy = "^2.1.3"
+numpy = "^1.0.0"
```
stefanwagnerdev, line 35: I think now we would have an issue if a customer requires numpy 2?

stefanwagnerdev (changes requested):

pyproject.toml
langenhagen: want `numpy = ">1.0.0"`

stefanwagnerdev: I guess this would be fine.

langenhagen: Can, do, will do if you want to, but generally - supporting numpy 1 AND 2 at the same time is not a good call because that effecitvely leaves us with only the intersection of numpy 1 and 2 to play around, everything that is not supported in either is gonna break in the other. Lucky us is that if we stick to the non-deprecated functions, which we seem to do, from eyeballing at least, this is not an issue with numpy specifically, apart from some potential float64 / float32 datatype bollocks. Numpy is largely benign in this regard, but other than that, that's not a good thing to do. If you want me to, I'll do it anyhow, here.

stefanwagnerdev (approved):

PR #113, opened 2025-03-06
feat(RPS-1094): add opc ua functions

mahsumdemirwb:


nova/extentions/opcua/client.py
```diff
@@ -0,0 +1,278 @@
...
+    It supports a condition function that is used to determine when the subscription should be completed.
+    """
+
+    def __init__(self, condition: Callable[[Any], bool], print_received_messages=False):
+        super().__init__()
+        self._flag = asyncio.Event()
+        self._condition = condition
+        self._print_received_messages = print_received_messages
+
+    async def datachange_notification(self, node, val, data) -> None:
```
langenhagen, line 33: nit: are type hint specifications for `node`, `val`, `data` feasible? Since we're now working on an SDK, type hints are especially valuable for 3rd party users.

langenhagen (approved):
lgtm

Didn't side-by-side compare with what we have in wandelbrain atm. I guess most is copy paste and thus tested before, as well as smoke-tested her in this repo.

nova/extentions/opcua/client.py
mahsumdemirwb: for the first part, yes It is %99 copy paste, just some small comment and class name changes

for the second part, even tought this class is not exported from the package, I agree types are useful so added them

PR #117, opened 2025-03-14
fix(RPS-1217): State streaming

functionistic:
Brings back the state streaming to the Controller device.


langenhagen:
generally lgtm, 

2 remarks:

1. the renaming state_stream to stream_state looks a bit like an edit-war that starts here?
2. the file `tests/state_stream.py` is not really a (py-)test? maybe another place for that file would be more suitable?

langenhagen (approved):
see my comment above, besides lgtmm

PR #118, opened 2025-03-17
chore(RPS-1129): Motion group should accessible via id str

biering:


nova/core/nova.py
```diff
@@ -39,7 +39,6 @@ def __init__(
         password: str | None = None,
         version: str = "v1",
         verify_ssl: bool = True,
-        log_level: str = LOG_LEVEL,
```
langenhagen, line 42: nit: looks unrelated intentional?

langenhagen (approved):
one nit about I am unsure whether this is intentional, otherwise lgtm

nova/core/nova.py
langenhagen: :+1:

PR #120, opened 2025-03-17
chore(RPS-1310): verbosify action names

langenhagen:
Still have aliases to their short counterparts.

Also rename string literals that are not directly tied to logic to
remain unambiguous, e.g. in `run_ptp_benchmark.py`.

plus, some minor commits


biering (approved):
lgtm

PR #123, opened 2025-03-18
fix(RPS-1312): Limit TCP default velocity to 50mm/s

langenhagen:
Tested via wandelscript's CLI tool `ws` and a modified default script; 

Kicks in when `velocity()` is not specified in the Wandelscript program. When `velocity()` is set inside the program, the setting picks up accordingly.


biering (approved):
lgtm

mahsumdemirwb (approved):

PR #125, opened 2025-03-20
fix(RPS-155): Provide motion group / robot ID on `PlanTrajectoryFailed` exceptions

langenhagen:
Also add this in the error message and pretty string function. This way, downstream clients may display the motion group ID in order to denote the robot that caused the exception.

Tested via Wandelscript CLI. Looks like:
2025-03-20-11-28-50-clipboard
note the mention of `0@controller`. That's the new stuff


biering (approved):
lgtm

PR #132, opened 2025-04-02
fix/pose-model-validate

mahsumdemirwb:


biering:
Please add a doctest / unit test to test this implementation

biering (approved):

PR #136, opened 2025-04-08
feat(rps-1476): isolate api client usage to api gateway class

mahsumdemirwb:


nova/core/nova.py
```diff
@@ -83,30 +77,13 @@ class Cell:
     """A representation of a robot cell, providing high-level operations on controllers."""
 
     def __init__(self, api_gateway: ApiGateway, cell_id: str):
-        """
```
langenhagen, line 86: nit: again, why remove a docstring?

langenhagen: same below

langenhagen (commented):
thanks, this makes the v1/v2 hassle much more visible.

I got to wonder about those removed docstrings though. What's up with that?

nova/core/gateway.py
```diff
@@ -257,6 +243,314 @@ def username(self) -> str | None:
...
+
+    async def list_controller_io_descriptions(
+        self, *, cell: str = None, controller: str = None, ios: list[str] = []
+    ) -> list[wb.models.IODescription]:
+        response = await self.controller_ios_api.list_io_descriptions(
+            cell=cell, controller=controller, ios=ios
+        )
+        return response.io_descriptions
+
+    async def get_controller_io(
```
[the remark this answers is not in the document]
mahsumdemirwb: I like read_controller_io better
refactored

biering:
Is the empty client.py file intended?

nova/core/nova.py
```diff
@@ -174,120 +117,86 @@ async def add_virtual_robot_controller(
         timeout: int = 25,
         position: str | None = None,
     ) -> Controller:
-        """Add a virtual robot controller to the cell
```
langenhagen, line 177: would leave the old docstring here, unless it became outdated

langenhagen (approved):
thanks for undoing the docstring deletions! The PR's changeset is much smaller now :D .Only one thing about a last docstring but that's it from my side. Already approving

nova/core/nova.py
mahsumdemirwb: I moved the extensive docstring to gateway.py file because the logic is implemented there but I can also add it here, would you prefer to have it in both? or in just one of them? if just one, which one would you prefer?

I don't have a preference that's why I am asking :D

biering (approved):

PR #137, opened 2025-04-11
chore(RPS-1557): migrate to `uv`

langenhagen:
Steps to port nova from based on `poetry` to being based on  `uv`, including changes to the documentations, CI, everything.


.github/workflows/nova-dev.yaml
```diff
@@ -18,16 +18,16 @@ jobs:
           python-version: "3.10"
       - name: Set up Python environment
         run: |
-          pip install poetry
-          poetry install --extras "nova-rerun-bridge"
+          pip install uv
+          uv sync --extra "nova-rerun-bridge"
       - name: Check formatting with ruff
-        run: poetry run ruff format --check .
+        run: uv run ruff format --check .
```
stefanwagnerdev, line 24: might be better to run `uvx ruff`

Context:
Tools are installed into temporary, isolated environments when using uvx.

Note
If you are running a tool in a [project](https://docs.astral.sh/uv/concepts/projects/) and the tool requires that your project is installed, e.g., when using pytest or mypy, you'll want to use [uv run](https://docs.astral.sh/uv/guides/projects/#running-commands) instead of uvx. Otherwise, the tool will be run in a virtual environment that is isolated from your project.

If your project has a flat structure, e.g., instead of using a src directory for modules, the project itself does not need to be installed and uvx is fine. In this case, using uv run is only beneficial if you want to pin the version of the tool in the project's dependencies.

.gitignore
```diff
@@ -164,18 +164,16 @@ models
 nova_rerun_bridge/rerun_app/models
 nova_rerun_bridge/rerun_app/nova_rerun_bridge
 nova_rerun_bridge/rerun_app/pyproject.toml
-nova_rerun_bridge/rerun_app/poetry.lock
```
stefanwagnerdev, line 167: do you want to add the uv.lock?

.github/workflows/nova-dev.yaml
langenhagen: Tried both locally, `uv run` instead of `uvx`. ruff already gets installed like the other aux tools, thus using `uv run` instead of `uvx`.

README.md
```diff
@@ -357,14 +357,14 @@ async def setup_coordinated_robots():
...
+uv sync --extra "nova-rerun-bridge"
 ```
 
 ### Formatting
 
 ```bash
-poetry run ruff format
-poetry run ruff check --select I --fix
+uv run ruff format
+uv run ruff check --select I --fix
```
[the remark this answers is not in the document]
langenhagen: I assume this is a nitpick?

Honestly, I would keep thinks vanilla, so no. Would leave it here. This is an SDK for the open public, thus I opt to keep it here.

.gitignore
langenhagen: Done :+1:

README.md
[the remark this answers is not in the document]
stefanwagnerdev: Yes nit, feel free to ignore. 
Just for some thoughts: I assume "vanilla" ruff check will already behave differently as we already have some ruff settings. I would keep it consistent and collect all settings inside the pyproject.toml

stefanwagnerdev (approved):

biering (approved):
lgtm

mahsumdemirwb (approved):

PR #138, opened 2025-04-14
feat(RPS-1509): Introduce a way to run a standalone Python robot program via uv

biering:


pyproject.toml
```diff
@@ -39,6 +39,7 @@ APScheduler = { version = ">=3.11.0", optional = true }
 trimesh = { version = ">=4.5.3", optional = true }
 pyyaml = { version = ">5.3", optional = true }
 asyncua = "^1.1.5"
+docstring-parser = "^0.16"
```
langenhagen, line 42: oh god, I will have a helluva time merging/rebasing un-conflicting my port from poetry to uv when the lockfile changes :thisisfine:

scripts/create_json_schema.py
```diff
@@ -0,0 +1,75 @@
...
+
+    file_path = Path(sys.argv[1])
+
+    if not file_path.exists():
+        print(f"File not found: {file_path}")
+        sys.exit(1)
+
+    try:
+        generate_schema_from_file(file_path)
+    except Exception as e:
```
langenhagen, line 69: nit: don't know if this try/except is actually helpful. I assume it's well-meaning, but basically, you're just nerfing the exception and exiting; just letting the exception throw without the try/except would basically cause similar behavior, but also show the stack trace.

langenhagen (approved):
lgtm, still weary about that changed dep and lockfile for my other uv-relater mr. because of the merge-conflicts

pyproject.toml
biering: I've resolved the merge conflicts

scripts/create_json_schema.py
```diff
@@ -0,0 +1,71 @@
...
+    # Find the main function
+    func_name = find_main_function(file_path)
+    if not func_name:
+        raise ValueError(f"No main function found in {file_path}")
+
+    # Get the function from the module
+    main_func = getattr(module, func_name)
+
+    # Generate the schema
+    schema = main_func.json_schema
```
functionistic, line 44: This will fail if `main_func` was not decorated.

functionistic (approved):

scripts/create_json_schema.py
biering: Added a check in the validate function to ensure its decorated

PR #139, opened 2025-04-15
feat(rps-1560): add support for pysical controllers

mahsumdemirwb:
Adding support for adding physical controllers and ensuring physical controllers.
Adding new helper functions for easy creation of configuration parameters


nova/core/nova.py
```diff
@@ -202,3 +253,123 @@ async def get_robot_cell(self) -> RobotCell:
         """
         controllers = await self.controllers()
         return RobotCell(timer=None, **{controller.id: controller for controller in controllers})
+
+
+def abb_controller(
+    name: str, controller_ip: str, egm_server_ip: str, egm_server_port: str
```
langenhagen, line 259: nit: I see you use the xxx_port variables of type `str`. wouldn't be type `int` more fitting? both in the type hints, doc strings and in the other occurences of _port in that PR.

Looked ad implementation of `AbbControllerEgmServer` and it seems to me port is a pydantic `StrictInt`, even.
```
class AbbControllerEgmServer(BaseModel):
    """
    The EGM server runs inside of the cell, thus its IP must be in the same network as the 'controller_ip'
    """ # noqa: E501
    ip: StrictStr
    port: StrictInt  # <<< !
```

langenhagen (commented):
largely lgtm, aside from the nit. would be happy about a short rationale and run-down in the PR's description. I read the code and think I grasp it, but would be happy about some text to double check, i.e., be on the safe side I understand the usage correctly.

nova/core/nova.py
```diff
@@ -202,3 +253,123 @@ async def get_robot_cell(self) -> RobotCell:
         """
         controllers = await self.controllers()
         return RobotCell(timer=None, **{controller.id: controller for controller in controllers})
+
+
+def abb_controller(
```
biering, line 258: You could further reduce the redundant logic by:

```python
def _build_controller(name: str, controller_obj: Any) -> api.models.RobotController:
    """
    Helper to wrap any given controller object into a standard RobotController.
    """
    return api.models.RobotController(
        name=name,
        configuration=api.models.RobotControllerConfiguration(controller_obj),
    )


def abb_controller(name: str, controller_ip: str, egm_server_ip: str, egm_server_port: str) -> api.models.RobotController:
    """
    Create an ABB controller configuration for a physical robot.
    """
    abb_config = AbbController(
        controller_ip=controller_ip,
        egm_server=api.models.AbbControllerEgmServer(
            ip=egm_server_ip,
            port=egm_server_port,
        ),
    )
    return _build_controller(name, abb_config)
    
...
```

examples/01_basic.py
```diff
@@ -10,17 +10,20 @@
 
 import asyncio
 
-from nova import Nova
+from nova import Nova, virtual_controller
```
biering, line 13: I'm not a fan of the import structure with this.

Maybe something like this would be more descriptive:

```python
from nova.controllers import abb_controller, kuka_controller, ...

# or

from nova import controllers

...

abb_controller = controllers.abb_controller(...)
```

mahsumdemirwb:
> largely lgtm, aside from the nit. would be happy about a short rationale and run-down in the PR's description. I read the code and think I grasp it, but would be happy about some text to double check, i.e., be on the safe side I understand the usage correctly.

Thanks for the review! I’ve added a brief rationale and summary to the PR description—let me know if anything’s unclear or if you’d like more detail.

nova/core/nova.py
mahsumdemirwb: Yes good catch! mypy also catched these and I am changing them, I guess I read StringInt as String :D

nova/core/nova.py
mahsumdemirwb: This looks much more readable, will do that

examples/01_basic.py
mahsumdemirwb: I aggree with the "code smell" ( https://martinfowler.com/bliki/CodeSmell.html) 
if we are going to create a new package strcuture, here is my suggestion:

create a new package called cell and move both these new functions and also the cell class there. I like keeping realted things closer and in this case these functions are there to support add_controller and ensure_controller functions of the cell

the nova.core package is growing and I think this will make it smaller

the code will still look like this:
```python
from nova.cell import Cell, virtual_controller, ...
```

because ruff wants to have a single line for everything imported from a package

If we really don't want this than we can go one level more and have
```python
from nova.cell import Cell
from nova.cell.controllers import virtual_controller, yaskawa...

```

This is still not a full solution to what you don't want to do(importing class and functions in a single line), because if we will add a class to the controller package then we will end up in the same situation :D

nova/core/nova.py
langenhagen: > StringInt

`StrictInt` /ftfy ^^

https://docs.pydantic.dev/2.3/api/types/#pydantic.types.StrictInt

mahsumdemirwb: :D :D thanks for pointing out to the docs, time to learn I guess :D

examples/01_basic.py
biering: I'm fine with having it in `cell`

nova/core/gateway.py
```diff
@@ -393,29 +382,19 @@ async def wait_for_controller_ready(self, cell: str, name: str, timeout: int = 2
...
-                    return
-
-            logger.info(f"Waiting for {cell}/{name} controller availability")
-            await asyncio.sleep(1)
             controller = await self.get_controller_instance(cell=cell, name=name)
+            if controller is None:
+                logger.info(f"Controller not found: {cell}/{name}")
+            elif controller.has_error:
+                logger.error(
+                    f"Controller has error: {cell}/{name}, error: {controller.error_details}"
```
functionistic, line 391:
```suggestion
                    f"Controller {cell}/{name} has error: {controller.error_details}"
```

nova/core/nova.py
```diff
@@ -202,3 +253,115 @@ async def get_robot_cell(self) -> RobotCell:
...
+    Args:
+        controller_ip (str): The IP address of the ABB robot.
+        egm_server_ip (str): The IP address of the EGM server.
+        egm_server_port (str): The port of the EGM server.
+    """
+    abb_config = AbbController(
+        controllerIp=controller_ip,
+        egmServer=api.models.AbbControllerEgmServer(ip=egm_server_ip, port=egm_server_port),
+    )
+
```
functionistic, line 287:
```suggestion
```

nova/core/nova.py
```diff
@@ -202,3 +253,115 @@ async def get_robot_cell(self) -> RobotCell:
...
+) -> api.models.RobotController:
+    """
+    Create a KUKA controller configuration for a physical robot.
+    Args:
+        controller_ip (str): The IP address of the KUKA robot.
+        controller_port (str): The port of the KUKA robot.
+        rsi_server_ip (str): The IP address of the RSI server.
+        rsi_server_port (str): The port of the RSI server.
+    """
+
```
functionistic, line 312:
```suggestion
```

nova/core/nova.py
```diff
@@ -202,3 +253,115 @@ async def get_robot_cell(self) -> RobotCell:
...
+    kuka_config = KukaController(
+        controllerIp=controller_ip,
+        controllerPort=controller_port,
+        rsiServer=api.models.KukaControllerRsiServer(ip=rsi_server_ip, port=rsi_server_port),
+    )
+
```
functionistic, line 318:
```suggestion
```

nova/core/nova.py
```diff
@@ -115,46 +121,91 @@ async def add_virtual_robot_controller(
...
+            cell=self._cell_id, name=robot_controller.name, timeout=wait_for_ready_timeout
+        )
+
+        return self._create_controller(robot_controller.name)
+
+    async def ensure_controller(
+        self,
+        robot_controller: api.models.RobotController,
+        add_timeout: int = 25,
+        waitfor_ready_timeout: int = 25,
```
functionistic, line 186: All of the timeouts seem arbitrary especially if they are all the same value. 
If they really should all be 25 seconds it hints at a general problem they address which then would be better described using a global constant for all of them instead of the literal value 25

nova/core/nova.py
```diff
@@ -202,3 +253,115 @@ async def get_robot_cell(self) -> RobotCell:
...
+    | VirtualController
+    | YaskawaController,
+) -> api.models.RobotController:
+    """Helper function to wrap a controller configuration in a RobotController object."""
+    return api.models.RobotController(
+        name=name, configuration=api.models.RobotControllerConfiguration(controller)
+    )
+
+
+def abb_controller(
```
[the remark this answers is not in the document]
mahsumdemirwb: `Cell.add_controller` and `Cell.ensure_controller` needs data for the controller configuration, the data model I choosed for these method is `wandelbots_api_client.models.RobotController`

creating this model is not user friendly. these helpers make creating this model more friendly. This is my reasoning for having helper functions. 


a quick look without helpers and with helper:
```python
api.model.RobotController(
  name: name,
  configuration: api.model.RobotControllerConfiguration(
    KukaController(
        controllerIp=controller_ip,
        controllerPort=controller_port,
        rsiServer=api.models.KukaControllerRsiServer(ip=rsi_server_ip, port=rsi_server_port),
    )
  )
)
```


```python
kuka_controller(name, controller_ip, controller_port, rsi_server_ip, rsi_server_port)
```

Then the next question is why not one helper but so many? Because each controller has different data that the user needs to pass:

kuka has rsi server related parameters
abb has egm server related parameters
virtual controller has no server but controller type related parameters

combining all in a single function would end up with too many optional parameters
or I could create a new data model but then I would end up doing mapping between our data models and auto generated data models which is something we don't want to do

nova/core/nova.py
mahsumdemirwb: These are indeed arbitrary, I just took them from the previous implementation.
The only decision I took in this MR is to seperate add_timeout and wait_for_ready_timeout and then I sticked to the previous arbitrary number.

About the need of timeout:
add_timeout is needed because it is passed down to the API
wait_for_ready_timeout is needed because we want to wait until the controller is ready before returning from the function and we need a timeout option for the user, we can't wait indefinetly

examples/01_basic.py
mahsumdemirwb: okay, will create a new package and move the Cell there

biering (approved):
lgtm

nova/core/nova.py
functionistic: I'm not saying that these timeouts make no sense. I'm saying, if all of them are the same, make a constant with a descriptive name, which indicates what the timeout is about.

nova/core/nova.py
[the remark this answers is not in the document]
functionistic: I didn't see the flattening of the parameters done by the convenience functions. That's the benefit I was missing.

PR #145, opened 2025-04-28
feat: add a example on how to serialize a program

stefanwagnerdev:
Request is to save and load a program. This example shows on how to serialize the joint trajectory and actions to reuse them later.


nova/actions/base.py
```diff
@@ -12,3 +13,49 @@ def serialize_model(self):
...
+    def _find_action_class_by_type(cls, action_type: str):
+        """Dynamically find an action class that has the given type literal value"""
+        # First, import all relevant modules to ensure classes are defined
+        from nova.actions import io, motions
+
+        # Look for the action class in all submodules
+        for module in [motions, io]:
+            for name, obj in inspect.getmembers(module):
+                # Check if this is a class that inherits from Action
+                if (
```
langenhagen, line 41: nit: consider `all()` instead of `if ... and ... and ... and ...` its often shorter and better readable

langenhagen: nit: on further look, I'd propose:
```
if all([
    isinstance(obj, type),
    issubclass(obj, Action),
    obj is not Action,
    "type" in getattr(obj, "model_fields", {}),
]):
    ...
 ```
    
 gets you rid of the inspect thing and and itself a bit less lines
 
 this code is likely to be copied by random users

**Review by langenhagen — APPROVED** [review]
lgtm apart from a nit

File: examples/11_serialize_program.py [hunk 2]
```diff
@@ -0,0 +1,97 @@
...
+                cartesian_ptp(target_pose @ [50, 0, 0, 0, 0, 0]),
+                joint_ptp(home_joints),
+                cartesian_ptp(target_pose @ (50, 100, 0, 0, 0, 0)),
+                joint_ptp(home_joints),
+                cartesian_ptp(target_pose @ Pose((0, 50, 0, 0, 0, 0))),
+                joint_ptp(home_joints),
+            ]
+
+        # you can update the settings of the action
+        for action in actions:
```
biering, line 55: NIT: To make sure that the user understands the intention of the example I would reduce it to the minimum and remove everything that is not really needed to show the serialize use case (e.g. setting the velocity, removing some actions etc.)

nova/actions/base.py
```diff
@@ -12,3 +13,49 @@ def serialize_model(self):
     @abstractmethod
     def is_motion(self) -> bool:
         """Return whether the action is a motion"""
+
+    @classmethod
+    def create_from_dict(cls, data: dict) -> "Action":
```
biering, line 18: Since our integration tests are rather slow it would be great to have some unit tests for this functionality.

functionistic (approved):

langenhagen:
@functionistic, @stefanwagnerdev , ah true, I revise my prior statement

examples/11_serialize_program.py
stefanwagnerdev: removed

nova/actions/base.py
stefanwagnerdev: I added a simple unit test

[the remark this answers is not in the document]
stefanwagnerdev: thanks

stefanwagnerdev:
The deserialization actually didn't contained the target and motion settings. I added them but the base Action class now contains a lot of mapping code. Maybe someone with more pydandic experience can improve that? @biering  
The issue I see is that the target field is overloded with tuble and pose...

PR #148, opened 2025-05-02
fix(RPS-1595)!: tcp_pose is returned with the correct TCP reference now

functionistic:
The `MotionGroupStateResponse` of the `MotionGroupInfosApi` has two places where it provides `tcp_pose` information.
The one inside of the `state` field may or may not be the current selected TCP of the robot (the description is contradictory there). The other is only present if a TCP was given to the `MotionGroupInfosApi.get_current_motion_group_state()` call and then is at the top level of the response object giving the TPC pose of the provided TCP.
`MotionGroup.get_state()` wrongly always used `tcp_pose` field inside of `state`. Now it uses the other one if a TCP was provided, mirroring the underlying behaviour, but uses `state.tcp_pose` if no TCP was given.

### DIFF
--- a/nova/core/motion_group.py
+++ b/nova/core/motion_group.py
@@ -424,13 +424,13 @@ class MotionGroup(AbstractRobot):
         """
         Returns the motion group state.
         Args:
-            tcp (str | None, optional): The reference TCP for the cartesian pose part of the robot state. Defaults to None.
+            tcp (str | None): The reference TCP for the cartesian pose part of the robot state. Defaults to None.
                                         If None, the current active/selected TCP of the motion group is used.
         """
         response = await self._api_gateway.get_motion_group_state(
             cell=self._cell, motion_group_id=self.motion_group_id, tcp=tcp
         )
-        pose = Pose(response.tcp_pose if response.tcp_pose is not None else response.state.tcp_pose)
+        pose = Pose(response.tcp_pose or response.state.tcp_pose)
         return RobotState(pose=pose, joints=tuple(response.state.joint_position.joints))
 
     async def joints(self) -> tuple:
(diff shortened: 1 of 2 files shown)


langenhagen (approved):
thanks for the description, makes sense!

notwendige message: seems that (unit-) tests are missing!

I'll approve nonetheless in order to unblock everything

nova/core/motion_group.py
```diff
@@ -421,14 +421,20 @@ async def stop(self):
             logger.debug(f"No motion to stop for {self}: {e}")
 
     async def get_state(self, tcp: str | None = None) -> RobotState:
+        """
+        Returns the motion group state.
+        Args:
+            tcp (str | None, optional): The reference TCP for the cartesian pose part of the robot state. Defaults to None.
```
biering, line 427:
```suggestion
            tcp (str | None): The reference TCP for the cartesian pose part of the robot state. Defaults to None.
```

nova/core/motion_group.py
```diff
@@ -437,6 +443,12 @@ async def joints(self) -> tuple:
         return state.joints
 
     async def tcp_pose(self, tcp: str | None = None) -> Pose:
+        """
+        Returns the current TCP pose of the motion group.
+        Args:
+            tcp (str | None, optional): The reference TCP for the returned pose. Defaults to None.
```
biering, line 449:
```suggestion
            tcp (str | None): The reference TCP for the returned pose. Defaults to None.
```

nova/core/motion_group.py
```diff
@@ -421,14 +421,20 @@ async def stop(self):
...
+            tcp (str | None, optional): The reference TCP for the cartesian pose part of the robot state. Defaults to None.
+                                        If None, the current active/selected TCP of the motion group is used.
+        """
         response = await self._api_gateway.get_motion_group_state(
             cell=self._cell, motion_group_id=self.motion_group_id, tcp=tcp
         )
-        return RobotState(
-            pose=Pose(response.state.tcp_pose), joints=tuple(response.state.joint_position.joints)
-        )
+        pose = Pose(response.tcp_pose if response.tcp_pose is not None else response.state.tcp_pose)
```
biering, line 433:
```suggestion
        pose = Pose(response.tcp_pose or response.state.tcp_pose)
```

PR #151, opened 2025-05-06
chore: improved serialize layer of actions

biering:


nova/actions/base.py
```diff
@@ -1,105 +1,60 @@
...
-                    # Check if this class has the right type value
-                    if hasattr(obj, "type") and getattr(obj, "type") == action_type:
-                        return obj
+    def from_dict(cls, data: dict) -> Action:
+        """
+        Pick the correct concrete Action from the `_registry`
+        and let Pydantic validate against that class.
+
+        Examples:
+        #>>> from nova.types import MotionSettings
```
langenhagen, line 35: nit: accidental comment hashes `#` ?

langenhagen (approved):
lgtm! skimmed over the changes, looks like a refactor, much code went away, nothing to complain, except for that 1 little nit.

nova/actions/base.py
biering: yes, needs to be removed

functionistic (approved):

PR #157, opened 2025-05-12
fix(RPS-1380): Upgrade pydantic

langenhagen:
Pydantic is the main bottleneck of our import times. The Pydantic team is aware of the performance issues and its recent 3.11 release was mainly focused on build performance. Upping its version effectively reduces our nova's import times. Since pydantic relies on only few harmless other libraries, I deem this narrowing of 3rd party dependencies to our SDK OK.


functionistic (approved):

langenhagen:
Tests seem to fail because of:

```
    from nova.core.robot_cell import RobotCell
E   ModuleNotFoundError: No module named 'nova.core.robot_cell'
```

seems like a legit error, but entirely unrelated to the changes in this PR. Thus, will merge notwithstanding this error.

PR #153, opened 2025-05-07
feat(RPS-1615): implemented program runner for Python program

biering:
- added abstract ProgramRunner class to introduce a general runtime concept in the SDK
- implemented UVProgramRunner based on the ProgramRunner to run uv standalone Python programs
- added tests for nova runtime

PR #158, opened 2025-05-13
chore: Towards passing tests

langenhagen:
Fix a missing `ExceptionGroup`s under Python 3.10; we seem to use them since commit 

`0838520 feat(RPS-1615): implemented program runner for Python program (#153)`

Also fix a broken import for `RobotCell`.

There are other issues related to recently added `TestProgramRunner`. I refrain from touching them since they've seen to be put there deliberately just a few days ago


biering (approved):
lgtm now

PR #169, opened 2025-05-23
feat(RPS-1600,  NDX-17): introducing cycle events, measurement and NATS propagation thereof

functionistic:
This introduces the `Cycle` class as a tool for the SDK user to track process cycles and cycle times. It also allows the user to write event handlers for the events generated by `Cycle`. These can be used to attach own actions/functions to the cycle events triggered by `Cycle`.


biering (approved):
lgtm

PR #173, opened 2025-05-27
feat(runner): added result field

biering:


langenhagen (approved):
lgtm, thanks for the easy pr :D

PR #176, opened 2025-06-02
fix: skip valid token check if a username and password is set

stefanwagnerdev:


biering:
The issue is now:
```
wandelbots_api_client.exceptions.NotFoundException: (404)
Reason: Not Found
HTTP response headers: <CIMultiDictProxy('Date': 'Mon, 02 Jun 2025 12:45:11 GMT', 'Content-Type': 'application/json', 'Content-Length': '74', 'Connection': 'keep-alive', 'Access-Control-Allow-Headers': 'DNT,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,Authorization', 'Access-Control-Allow-Methods': 'GET, PUT, POST, DELETE, PATCH, OPTIONS', 'Access-Control-Allow-Origin': '*', 'Strict-Transport-Security': 'max-age=31536000; includeSubDomains', 'Access-Control-Allow-Credentials': 'true', 'Access-Control-Max-Age': '1728000')>
HTTP response body: {"code":5,"message":"IO with id 'tool_out%5B0%5D' not found","details":[]}
```

This was the reason I did not merge the NOVA version update branch before. So it seems there is an issue with the I/Os in NOVA. Could also be a bug in NOVA.

biering (approved):
lgtm

PR #216, opened 2025-07-01
Feature/transparent rerun integration

stefanwagnerdev:


nova/program/function.py
```diff
@@ -269,13 +281,18 @@ class Output(RootModel[T]):  # pylint: disable=redefined-outer-name
     return input, output
 
 
-def program(name: str | None = None, preconditions: ProgramPreconditions | None = None):
+def program(
+    name: str | None = None,
+    preconditions: ProgramPreconditions | None = None,
+    viewer: Any | None = None,
```
biering, line 287: Would it make sense to have a ViewerProtocol here, to make it possible to create other viewers?

It would be fine for me to do this separately.

biering (approved):
lgtm, great work!

nova/program/function.py
stefanwagnerdev: Yes makes total sense. 
But yes, this PR is already to big

PR #217, opened 2025-07-02
feat(NDX-104): introduced Novax app framework

biering:


mahsumdemirwb (approved):

novax/api/programs.py
```diff
@@ -0,0 +1,94 @@
+from fastapi import APIRouter, Body, Depends, HTTPException, Path
+
+from novax.api.dependencies import get_program_manager
+from novax.program_manager import ProgramDetails, ProgramManager, ProgramRun, RunProgramRequest
+
+router = APIRouter(prefix="/programs", tags=["programs"])
+
+
+@router.get("", operation_id="getPrograms", response_model=list[ProgramDetails])
+async def get_programs(program_manager: ProgramManager = Depends(get_program_manager)):
```
functionistic, line 10:
```suggestion
async def list_programs(program_manager: ProgramManager = Depends(get_program_manager)):
```

novax/novax.py
```diff
@@ -0,0 +1,99 @@
...
+        # Replace the dependency function on the FastAPI app
+        app.dependency_overrides[get_program_manager] = get_program_manager_override
+
+        # Include the programs router
+        app.include_router(programs_router)
+
+        return app
+
+    def run(self):
+        pass
```
functionistic, line 99: What's this for?

biering: forgot to delete. thx

pyproject.toml
```diff
@@ -59,6 +59,11 @@ wandelscript = [
     "geometricalgebra>=0.1.3,<0.2",
     "typer[all]>=0.12,<0.16",
 ]
+novax = [
```
mahsumdemirwb, line 62: @biering @functionistic should we add wandelscript as dependency to novax?
because when I add novax to my project and try to register a wandelscript function app start up fails

since register_program functions support wandelscript by default, IMO, when you install novax, you should get wandelscript as well

mahsumdemirwb: or maybe something like this?
try:
  import wandelscript
except:
   raise Exception('to use wandelscript programs, please install the package first")

biering: true, good point

biering: image
seems like this is not so easy without duplicating the wandelscript deps.
We could also think about making Wandelscript not optional

mahsumdemirwb: sounds good to me

functionistic (approved):

novax/program_manager.py
```diff
@@ -0,0 +1,214 @@
...
+import wandelscript
+from nova import Nova
+from nova.program.function import Function
+from nova.program.runner import ExecutionContext, Program, ProgramRun, ProgramRunner, ProgramType
+
+
+class ProgramSource(Protocol):
+    """Protocol for program sources that can be registered with ProgramManager"""
+
+    def get_programs(self, program_manager: "ProgramManager") -> AsyncIterator[Function | Path]:
```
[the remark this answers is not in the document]
mahsumdemirwb: strong +1, I would love to have strong support for foreign functions, I think they are super useful

novax/novax.py
```diff
@@ -0,0 +1,96 @@
...
+            title: The title of the API
+            version: The version of the API
+
+        Returns:
+            FastAPI: The configured FastAPI application
+        """
+        if self._app is not None:
+            return self._app
+
+        self._app = FastAPI(
```
mahsumdemirwb, line 66: @biering I have a small question about the this.
as I understand, when we create the FastApi like this, it gives swagger UI under the /docs where our users can use it.

should we instead give spotlight UI? because this is what the core team gives for Nova API, IMO, this will give better, more unified user experience.

here is an example code I found:

```python
@app.get("/", summary="Opens the Stoplight UI", response_class=HTMLResponse)
async def root():
    # One could serve a nice UI here as well. For simplicity, we just redirect to the Stoplight UI.
    return f"""
    <!doctype html>
        <html lang="en">
          <head>
            <meta charset="utf-8">
            <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
            <title>Elements in HTML</title>
            <!-- Embed elements Elements via Web Component -->
            <script src="https://unpkg.com/@stoplight/elements/web-components.min.js"></script>
            <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements/styles.min.css">
          </head>
          <body>

            <elements-api
              apiDescriptionUrl="{BASE_PATH}/openapi.json"
              router="hash"
              layout="sidebar"
              tryItCredentialsPolicy="same-origin"
            />

          </body>
    </html>
    """
```

functionistic: The template app(s) also does it like this.

novax/program_manager.py
```diff
@@ -0,0 +1,214 @@
...
+        if inspect.iscoroutine(result):
+            result = await result
+
+        return result
+
+
+class ProgramDetails(BaseModel):
+    program_id: str
+    created_at: str
+    updated_at: str
```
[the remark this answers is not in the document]
biering: replaced with datetime

PR #225, opened 2025-07-10
feat: provide nova to program & update novax API

biering:
This MR lets the decorator provide the nova object to the program as a first parameter. With that the user does not need to do `async with Nova() as nova` in every program. The downside is that the first parameter always needs to be nova: Nova


biering:
I think this introduces too much complexity now. I would not merge it for now since we have a more clean customer benefit

PR #234, opened 2025-07-16
feat: added deregister_program to novax

biering:
* added unit tests to program manager
* renamed run_program to start_program
* added deregister_program
* removed old Program & ProgramType model


mahsumdemirwb (approved):

novax/program_manager.py
```diff
@@ -144,6 +149,18 @@ def register_program(self, program: Program) -> str:
...
 
+    def deregister_program(self, program_id: str):
+        """
+        Deregister a program from the program manager
+
+        Args:
+            program_id: The ID of the program to deregister
+        """
+        if program_id not in self._programs:
+            raise ValueError(f"Program {program_id} not found")
```
functionistic: Maybe prefer idempotency?
Or framed differently: What does the user gain from this exception being raised?

tests/novax/test_program_manager.py
```diff
@@ -0,0 +1,255 @@
...
+def test_deregister_nonexistent_program():
+    manager = ProgramManager(robot_cell_override=SimulatedRobotCell())
+
+    try:
+        manager.deregister_program("nonexistent_program")
+        assert False, "Expected ValueError to be raised"
+    except ValueError as e:
+        assert "Program nonexistent_program not found" in str(e)
+
+
```
functionistic, line 91: Or test that nothing happens.
```suggestion
```

tests/novax/test_program_manager.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.mark.asyncio
+async def test_stop_wrong_program():
+    manager = ProgramManager(robot_cell_override=SimulatedRobotCell())
+
+    # Register programs
+    manager.register_program(long_running_program)
+    manager.register_program(simple_program)
+
+    # Start the first program
+    await manager.start_program("long_running_program")
```
functionistic, line 174: It only occurred here to me, that we might want to get the program id (name) from the program object instead of using the string here. The string is hard to rename later. 
But for tests this is rather nitpicky.

functionistic (approved):

PR #235, opened 2025-07-16
feat: add program store

mahsumdemirwb:


nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,142 @@
...
+from nats.js import JetStreamContext
+from nats.js.api import KeyValueConfig
+from nats.js.client import KeyValue
+from nats.js.errors import KeyNotFoundError as KvKeyError
+from nats.js.errors import NotFoundError, NoKeysError
+from nova.core.logging import logger
+
+from pydantic import BaseModel, ValidationError
+
+from nova.program.function import Program
```
functionistic, line 16: Import order is off.

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,142 @@
...
+
+# generally people use NATS_SERVERS, app store uses NATS_BROKERS
+NATS_SERVERS = config("NATS_SERVERS", default=None, cast=str)
+if not NATS_SERVERS:
+    NATS_SERVERS = config("NATS_BROKERS", default="nats://nats.wandelbots.svc:4222", cast=str)
```
functionistic, line 21: We don't want to have multiple names for the same thing. NATS_BROKERS is just wrong.
image

functionistic:
```suggestion
```

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,142 @@
...
+        ):
+        self.model_class = model_class
+        self.kv_bucket_name = nats_bucket_name
+
+        self.nats_client_config = nats_client_config or {}
+        self.kv_config = nats_kv_config or {}
+
+        self.nc: NATS | None = None
+        self.js: JetStreamContext | None = None
+        self.kv: KeyValue | None = None
```
functionistic, line 45: I don't believe all of these need to be public.

functionistic:
How is this supposed to be used?

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,142 @@
...
+            logger.info(f"Using already existing KV bucket '{self.kv_bucket_name}'")
+        except NotFoundError:
+            if not self.kv_config:
+                raise Exception(f"The bucket with name: {self.kv_bucket_name} doesn't exists and kv_config is not provided, so it cannot be created.")
+
+            logger.info(f"No bucket found, creating new KV bucket '{self.kv_bucket_name}'")
+            kv = await self.js.create_key_value(config=self.kv_config)
+        return kv
+
+    async def shutdown(self) -> None:
```
functionistic, line 82: Who calls this?

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,142 @@
...
+    
+    def __init__(
+            self,
+            model_class: Type[T],
+            nats_bucket_name: str,
+            nats_client_config: dict | None = None,
+            nats_kv_config: KeyValueConfig | None = None,
+        ):
+        self.model_class = model_class
+        self.kv_bucket_name = nats_bucket_name
```
functionistic, line 38: Is it a `nats_bucket` or a `kv_bucket`?

functionistic (changes requested):
How is this supposed to be used?
A general question is whether we want to have multiple places in the SDK where NATS connections are being opened.

nova/program/nats_key_value_store.py
mahsumdemirwb: good point, will make private

mahsumdemirwb:
> How is this supposed to be used? A general question is whether we want to have multiple places in the SDK where NATS connections are being opened.

Currently, the plan is to use this in wandelengine and Novax.
Novax needs to put/delete programs and wandelengine needs to query it, get_all.
They need to integrate ProgramStore to the appropriate parts of the FastAPI, most probably lifespan.

After a couple of iteration, I found it more useful to have a generic key value store with a pydantic model. So that it is easy to change the model when we have the final version and also implement new key-value store based solutions to our architecture.

Note: Also, ProgramStore will use pydantic model from auto generated client when it is ready.

nova/program/nats_key_value_store.py
mahsumdemirwb: I created this ticket: https://wandelbots.atlassian.net/browse/NC-2086
This is not a blocker right? we should fix this in parallel with the core team and move on

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,221 @@
...
+if not NATS_SERVERS:
+    NATS_SERVERS = config("NATS_BROKERS", default="nats://nats.wandelbots.svc:4222", cast=str)
+
+# NATS_TOKEN is optional, but needed when you connect remotely to a NATS server
+NATS_TOKEN = config("NATS_TOKEN", default=None, cast=str)
+
+T = TypeVar("T", bound=BaseModel)
+
+
+class KeyValueStore(Generic[T]):
```
biering, line 27:
```suggestion
class NatsKeyValueStore(Generic[T]):
```

nova/program/__init__.py
```diff
@@ -1,4 +1,16 @@
-from nova.program.function import ProgramPreconditions, program
+from nova.program.function import Program, ProgramPreconditions, program
+from nova.program.nats_key_value_store import KeyValueStore
```
biering, line 2:
```suggestion
from nova.program.nats_key_value_store import NatsKeyValueStore
```

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,221 @@
...
+
+        self._nats_client_config = nats_client_config or {}
+        self._nats_kv_config = nats_kv_config
+
+        self._nc: NATS | None = None
+        self._js: JetStreamContext
+        self._kv: KeyValue
+        self._bucket_lock = asyncio.Lock()
+
+    async def connect(self) -> None:
```
biering, line 104:
```suggestion
    async def connect(self):
```

nova/program/nats_key_value_store.py
```diff
@@ -0,0 +1,221 @@
...
+            except NotFoundError:
+                if not self._nats_kv_config:
+                    raise RuntimeError(
+                        f"Bucket {self._nats_bucket_name} missing and no kv_config supplied"
+                    )
+                self._kv = await self._js.create_key_value(self._nats_kv_config)
+
+        return self._kv
+
+    async def put(self, key: str, model: T) -> None:
```
biering, line 163:
```suggestion
    async def put(self, key: str, model: T):
```

nova/program/__init__.py
```diff
@@ -1,4 +1,16 @@
...
 from nova.program.runner import ProgramRunner
 
-__all__ = ["ProgramRunner", "program", "ProgramPreconditions"]
+# ProgramStore = KeyValueStore[Program] would be better but python doesn't support this
+# when I do store = ProgramStore() the __orig_class__ is not available in the __init__
+# my reseach say's python doesn't capture the type argument when I do this
+
+
+# TODO: change the Program with wandelbots_api_client.v2.models.Program
+class ProgramStore(KeyValueStore[Program]):
```
biering, line 11: Minor: Maybe put this under /program/store.py. I don't like if some code is in the files and some code is in the __init__

PR #236, opened 2025-07-21
feat(NDX-174): implemented TrajectoryBuilder

biering:
Added trajectory builder to easily fine-tune motion settings

```python
# The trajectory builder is a context manager that can be used to build a trajectory with fine-grained control over the settings
t = TrajectoryBuilder(settings=normal)

# First move to the home position, since we passed "slow" settings to the motion it will override the one from the trajectory builder
t.move(joint_ptp(home_joints, settings=slow))

# It's possible to use the sequence method to add multiple actions to the trajectory. Since no settings are specified it takes
#   the settings from the trajectory builder
t.sequence(
    [
        cartesian_ptp(target_pose),
        joint_ptp(home_joints),
        cartesian_ptp(target_pose @ Pose((50, 0, 0, 0, 0, 0))),
        joint_ptp(home_joints),
        io_write(key="tool_out[0]", value=False),
    ]
)

# You can use the set(...) context manager to set settings for a block of actions
with t.set(settings=fast):
    t.move(cartesian_ptp(target_pose @ Pose((50, 100, 0, 0, 0, 0))))
    t.move(
        joint_ptp(home_joints, settings=slow)
    )  # Moving back to home with slow settings
    t.move(cartesian_ptp(target_pose @ Pose((0, 50, 0, 0, 0, 0))))

    t.move(joint_ptp(home_joints, settings=slow))
```

### DIFF
--- a/nova/program/function.py
+++ b/nova/program/function.py
@@ -133,9 +133,16 @@ class Program(BaseModel, Generic[Parameters, Return]):
             async with Nova() as nova:
                 cell = nova.cell()
                 for controller_id in controller_ids:
-                    await cell.delete_robot_controller(controller_id)
-                    self._log("info", f"Cleaned up controller with ID '{controller_id}'")
+                    try:
+                        await cell.delete_robot_controller(controller_id)
+                        self._log("info", f"Cleaned up controller with ID '{controller_id}'")
+                    except Exception as e:
+                        # WORKAROUND: {"code":9, "message":"Failed to 'Connect to Host' due the
+                        #   following reason:\nConnection refused (2)!\nexception::CommunicationException: Configured robot connection is not reachable.", "details":[]}
+                        # Log and suppress errors for individual controller cleanup
+                        self._log("error", f"Error cleaning up controller '{controller_id}': {e}")
         except Exception as e:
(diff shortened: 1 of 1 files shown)


nova/actions/__init__.py
```diff
@@ -30,4 +33,6 @@
     "wait",
     "MovementController",
     "MovementControllerContext",
+    "TrajectoryBuilder",
+    "TrajectoryMacher",
```
stefanwagnerdev, line 37: 😁

nova/actions/container.py
stefanwagnerdev, line 1: I think it would be nice if it also handles the collision free settings which are not using the motion settings

stefanwagnerdev (approved):

nova/actions/container.py
stefanwagnerdev: Sure not necessary to fix it in that PR. But as a user I expect that the collision free p2p behaves exactly the same as the other actions if I define a "slow" motion.

biering: Yes, but for the sake of getting more feedback from schaeffler faster I would merge it and tackle the collision free topic in an extra MR.

PR #257, opened 2025-08-12
fix: re-enabled Wandelscript unit tests & fixed them

biering:


nova/program/runner.py
```diff
@@ -83,11 +84,7 @@ class ProgramRunner(ABC):
     """
 
     def __init__(
-        self,
-        program_id: str,
-        args: dict[str, Any],
-        robot_cell_override: RobotCell | None = None,
-        simulate: bool = False,
+        self, program_id: str, args: dict[str, Any], robot_cell_override: RobotCell | None = None
```
stefanwagnerdev, line 87: simulate doc string is still there

functionistic (approved):

PR #272, opened 2025-09-03
Feature/added from euler util function

plexHD:
This pull request enhances the `Pose` class with the `from_euler` utility function. It allows the user to create cartesian `Poses` from Euler angles easily, by providing the position, Euler angles, convention ("xyz" by default), and the option to specify the angles in degrees (false by default). This makes creating `Poses` in the SDK a lot more intuitive, as the regular `Pose` constructor only accepts rotation vectors. Without any tools, these are very hard to visualize and make setting an orientation rather difficult.

Creating a Pose from Euler angles would look something like this:

`position = (100, 0, 0)`
`euler_angles_rad = (0, np.pi / 4, np.pi / 2)`
`new_pose = Pose.from_euler(position, euler_angles_rad, convention="xyz", degrees=False)`

### DIFF
--- a/examples/Euler_angles_Pose.py
+++ b/examples/Euler_angles_Pose.py
@@ -1,0 +1,69 @@
+"""
+Example: Perform relative movements with a robot.
+
+Prerequisites:
+- Create an NOVA instance
+- Set env variables (you can specify them in an .env file):
+    - NOVA_API=<api>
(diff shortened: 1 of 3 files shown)


biering (approved):
add new code to plan_and_execute.py example and minor changes

PR #290, opened 2025-09-29
feat: add auto mode management for controller and motion group usage

stefanwagnerdev:
Problem: Frequent MONITOR↔CONTROL mode switches cause expensive brake cycling and mode switching during trajectory execution.

**Solution:** 
- Add `Controller.set_default_mode()` and `get_current_mode()` methods
- `MotionGroup` automatically switches MONITOR→CONTROL on first execution
- Restores MONITOR mode on context exit
- Only switches when beneficial (respects existing CONTROL mode)

### DIFF
--- a/nova/core/controller.py
+++ b/nova/core/controller.py
@@ -16,7 +16,7 @@ class Controller(Sized, AbstractController, NovaDevice, IODevice):
     """
 
     class Configuration(NovaDevice.Configuration):
-        type: Literal["controller"] = "controller"  # type: ignore[assignment]
+        type: Literal["controller"] = "controller"
         id: str = "controller"
         cell_id: str
         controller_id: str


nova/core/motion_group.py
```diff
@@ -372,6 +398,21 @@ async def _execute(
...
+                logger.info(
+                    f"Motion group {self._motion_group_id}: Switching controller from MONITOR to CONTROL mode for execution session"
+                )
+                await self._controller.set_default_mode(RobotSystemMode.ROBOT_SYSTEM_MODE_CONTROL)
+                self._switched_from_monitor_to_control = True
+            else:
+                logger.info(
+                    f"Motion group {self._motion_group_id}: Controller already in CONTROL mode, no switch needed"
+                )
+
```
functionistic, line 415: I'd prefer it to be at the entrance of the motion group async context instead of the first execution.
This way the user can still control it and I think it solves the problem for most cases (also the EY one).
I see that there is no paramater to `open()` allowing this right now but are there any other reasons for not doing it like that?

stefanwagnerdev: The use case could be that users only want to listen to the robot configuration and not actively control the robot and free up the controls for other processes.

mahsumdemirwb: IMO, with the addition of features in the controller class
it is easy enough for a user to set the controller to the control mode and if they do that glitching will not happen

so my suggestion would be to not do this automatically for the user in motion group, especially because in v2 motion group activation concept is removed and maybe we could get rid of async context for the motion group all together and simplify user code

PR #301, opened 2025-10-14
feat: multi motion group example

adozzer:
Added an example on synchronous motion of several motion groups of one controller, using io for synchronisation

### DIFF
--- a/examples/multi_motion_group.py
+++ b/examples/multi_motion_group.py
@@ -1,11 +1,10 @@
 import asyncio
 from pathlib import Path
 
-import wandelbots_api_client.models as v1models
 from pydantic import TypeAdapter
 
 import nova
-from nova import run_program
+from nova import api, run_program
 from nova.cell import virtual_controller
 from nova.program import ProgramPreconditions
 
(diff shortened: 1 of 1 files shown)


examples/multi_motion_group.py
```diff
@@ -0,0 +1,97 @@
...
+        await controller.write(key=SYNC_IO_ID, value=True)
+
+    async with nova.Nova() as n:
+        cell = n.cell()
+        controller = await cell.controller("kuka")
+        robot = controller.motion_group(ROBOT_MOTION_GROUP_ID)
+        positioner = controller.motion_group(POSITIONER_MOTION_GROUP_ID)
+
+        # Load trajectories for both motion groups (same duration expected).
+        trajectories = load_joint_trajectories()
```
mahsumdemirwb: is it possible to calculate these trajectories using motion_group.plan function?
If it is possible what do you think of having some example poses and using them to show case the sync motion group robot movement?

mahsumdemirwb:
@adozzer
generally, looks good to me

but I think we need to take some decisions to fully integrate the start on io to the sdk
following topics are in my mind:
- should robot move to the start position before the IO is set? currently we move to the start position and than execute the trajectory
- should we integrate this to the trajectory builder? I think it will be good to add it there


@biering @functionistic

examples/multi_motion_group.py
```diff
@@ -0,0 +1,147 @@
...
+ROBOT_TCP_ID = "1"
+POSITIONER_TCP_ID = "0"
+
+ROBOT_MOTION_GROUP_ID = "0@kuka"
+POSITIONER_MOTION_GROUP_ID = "1@kuka"
+
+SYNC_IO_ID = "OUT#1"
+
+
+def move_forward_in_sync(
```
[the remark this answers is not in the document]
biering: Is it required in the SimulatedRobot? If not I would not add it there.

examples/multi_motion_group.py
```diff
@@ -0,0 +1,97 @@
+import asyncio
+from pathlib import Path
+
+import wandelbots_api_client.models as v1models
```
biering:
```suggestion
from nova import api
```

mahsumdemirwb (approved):

PR #320, opened 2025-10-28
feat: update rerun to 0.26, show rerun only locally, reruse nova instance in rerun NDX-370, NDX-367

stefanwagnerdev:


biering (approved):
lgtm

PR #321, opened 2025-10-29
fix: stop programs when lifespan goes down

mahsumdemirwb:
When web server goes down, we call stop on the program manager to not hang the server in case there a program running.

### DIFF
--- a/tests/novax/api/test_programs.py
+++ b/tests/novax/api/test_programs.py
@@ -1,9 +1,4 @@
-import asyncio
-
 from fastapi.testclient import TestClient
-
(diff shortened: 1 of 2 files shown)


tests/novax/api/test_programs.py
```diff
@@ -1,5 +1,10 @@
+import asyncio
+
 from fastapi.testclient import TestClient
 
+import nova
+from novax.novax import Novax
```
biering:
```suggestion
from novax import Novax
```

functionistic (approved):

PR #331, opened 2025-11-04
feat(NDX-43)!: upgrade Python SDK to v2

biering:
# Python SDK for NOVA API v2 Release Notes & Migration Guide
With this release we start using V2 apis Nova platform from python SDK.

## MotionSettings
- `position_zone_radius` is deprecated and replaced by `blending_radius`
- `min_blending_velocity` is deprecated and replaced by `blending_auto`
----

## Collision Free Motion Planning
- new parameter algorithm added. We use RRT as default. check `nova.actions.motions.collision_free` for more information
----

## Module restructuring
- Moved `motion_group.py`, `controller.py`, `io.py` and `movement_controllers.py` to `/cell` module
- Streamlined `nova.api` API access by using only `v2_pydantic` of the autogenerated API client
----

## Motion Group
- Added forward kinematics
- compare_collision_scenes, validate_collision_scenes, combine_trajectories moved to utils
- Added `get_model`
- Added `get_setup`
- Added `get_mounting`
 - Added `get_safety_collision_setup`
 - Added `get_default_collision_link_chain`
 - Added `tcp_offset`
 - movement_controller interface has changed as per Nova API.
 - movement logic changed, in V1, if you executed a trajectory without being on the start position we would first move to the start position and then execute your trajectory. Now what happens? check with @functionistic 
----

##  Others
- Nova class now exposes ApiGateeway with a field named `api`. Internal _api_client will be removed in a future release
----

## Breaking Changes
- `combine_trajectory` is removed from root `__init__.py`
- Changed `Pose.to_api_pose` to `Pose.to_api_model`
- Changed MotionGroup `.motion_group_id` to `.id`
- Changed Controller `.controller_id` to `.id`
- Upgraded minimum Python version to 3.11
- Removed nova.nats package. Nova class exposes library integration from nova.nats field.
- Removed Auth0 package and related environment variables
- Logger Changes:
  - LOGGER_NAME env var is removed. Top level logger is named “nova”. Each python sub-module logs to its own logger.
  - Removed deprecated nova.core.logger module
- NovaConfig:
   - make host in NovaConfig non-optional
   - remove deprecated username and password environment variables
   - host and NOVA_API parameter requires explict schema. ( http or https)
- When no motion settings provided, we use default motion settings with max velocity set to 50 and all else not set
- Cell.nats return nats library instance instead of NatsClient
-  nova/core/env_handler is remove
- ApiGateway now accepts NovaConfig as constuctor parameter
----

### DIFF
--- a/nova/viewers/rerun.py
+++ b/nova/viewers/rerun.py
@@ -311,7 +311,6 @@ class Rerun(Viewer):
                     else:
                         # TODO: handle collision free failed response
                         logger.warning("Collision free failed response not supported yet")
-                        pass
 
             # Log error information as text
             import rerun as rr


nova/core/io.py
```diff
@@ -15,8 +15,8 @@ class IOType(Enum):
 
 class IOValueType(Enum):
     IO_VALUE_ANALOG_INTEGER = "IO_VALUE_ANALOG_INTEGER"
-    IO_VALUE_ANALOG_FLOATING = "IO_VALUE_ANALOG_FLOATING"
-    IO_VALUE_DIGITAL = "IO_VALUE_DIGITAL"
+    IO_VALUE_ANALOG_FLOATING = "IO_VALUE_ANALOG_FLOAT"
+    IO_VALUE_DIGITAL = "IO_VALUE_BOOLEAN"
```
functionistic: Confusing diff between names and values.

PR #332, opened 2025-11-04
fix(NDX-403): wait_for_bool_io()

functionistic:
This pull request makes minor improvements to the `nova/core/gateway.py` file, focusing on import cleanup and a small fix to the use of an enum value.

- Imports and Backward Compatibility:
  * Cleaned up the import of `INTERNAL_CLUSTER_NOVA_API` to avoid duplicate imports and maintain backward compatibility.

- Enum Usage Fix:
  * Updated the `comparison_type` argument in the `wait_for_bool_io` method to use the `.value` of the `ComparisonType` enum, ensuring the correct value is passed.


biering (approved):
lgtm

PR #334, opened 2025-11-07
fix: update log level from warning to debug

mahsumdemirwb:
following warning messages show up in every import of the python package but they don't provide any meaningful information to the user. It only confuses them
This MR updates the log level from warning to debug.


biering (approved):
lgtm

PR #342, opened 2025-11-20
ndx-332-collision-free-for-v2

mahsumdemirwb:
# Content of the MR
- Adds inverse kinematics to motion_group.
- Adds utility to patch a motion group setup with motion settings
- Updates get_setup so when the API doesn't return link chain for the motion group, we add it
  - Open topic: What about TCP?
- Updates how tool_offset is passes to forward_kinematics
- Adds collision free action and integrates it with the planning.
   - Updates existing tests about split_batch_into_actions to be green.
   - When collision free action has pose as a target, we do inverse kinematics for the user with the same collision setup provided to the action.

# Tests
- Integration tests for inverse kinematics
- Integration tests for collision free planning
- Unit tests for utilities

# Important Points:
### Collision Setup Handling
- We expect that the user provides a proper collision setup. For this, user needs to do:

```python
setup = mg.get_setup()
# update the colliders or link_chains
# pass it to the other functions
mg.inverse_kinematics()

collision_free(collision_setup=...)
```

- When user doesn't provide a motion_group_setup, we use what `get_setup` returns.

### CollisionFreeMotion.to_api_model is missing
We need to create a `api.models.PlanCollisionFreeRequest` instance but this model requires a motion_group_setup and we do not have access to this data from CollisionFreeMotion

### inverse_kinematics api returns a complex data structure
what we get from the api is a list of list of list of floats, yes 3 lists.
- one list for each pose we do IK for
- one list for each solition found for that pose
- one list for the joint positions

This is definetly not easy to use.

### JointLimit creation looks weird
TODO: mahsum to add more details. 
The API model has couple of nested optional fields which arises the question of what if join limit for one is missing.

@functionistic @biering

### DIFF
--- a/tests/nova/cell/motion_group/test_collision_free_planning.py
+++ b/tests/nova/cell/motion_group/test_collision_free_planning.py
@@ -2,6 +2,7 @@ import pytest
 from nova import Nova
 from nova.cell import virtual_controller
 from nova.api import models
+from nova.cell.motion_group import MotionGroup
 from nova.types import Pose
 from nova.actions import collision_free
 from math import pi
@@ -93,7 +94,7 @@ async def test_collision_free_planning_with_pose_as_target(ur_mg):
 
 
 @pytest.mark.asyncio
-async def test_collision_free_planning_finds_no_solution(ur_mg):
+async def test_collision_free_planning_finds_no_solution(ur_mg: MotionGroup):
     """
     Tests that collision-free planning correctly identifies when no solution is possible.
     """
@@ -103,14 +104,17 @@ async def test_collision_free_planning_finds_no_solution(ur_mg):
             position=models.Vector3d(root=[0, 0, 0]),
             orientation=models.RotationVector(root=[0, 0, 0]),
         ),
+    
     )
-    collision_setup = models.CollisionSetup(colliders={"plane": plane})
+    setup = await ur_mg.get_setup("Flange")
+    default_collision_setup = setup.collision_setups.root["default"].model_copy()
+    default_collision_setup.colliders = {"plane": plane}
 
     with pytest.raises(Exception):
         await ur_mg.plan(
             start_joint_position=tuple(initial_joint_positions),
             actions=[
-                collision_free(target=Pose(700, 0, -10, 0, 0, 0), collision_setup=collision_setup)
+                collision_free(target=Pose(700, 0, -10, 0, 0, 0), collision_setup=default_collision_setup)
             ],
             tcp="Flange",
         )


nova/cell/motion_group.py
```diff
@@ -134,6 +149,60 @@ async def get_mounting(self) -> Pose | None:
...
+        Returns:
+            list[list[tuple[float, ...]]]: All found joint position solutions for each pose. The outer list corresponds to each pose,
+                                            and the inner list contains each valid joint configuration solution found for that pose.
+        """
+        motion_group_setup = await self.get_setup(tcp)
+        if collision_setup is not None and motion_group_setup.collision_setups is not None:
+            motion_group_setup.collision_setups.root["user"] = collision_setup
+
+        motion_group_description = await self._fetch_motion_group_description()
+        tcp_offset = (
```
biering: In the latest v2 branch you can use `await self.tcp_offset(tcp)` to get the pose offset

nova/cell/motion_group.py
```diff
@@ -145,7 +214,13 @@ async def forward_kinematics(self, joints: list[tuple[float, ...]], tcp: str) ->
 
         joint_positions = [api.models.DoubleArray(list(joint_config)) for joint_config in joints]
 
-        tcp_offset = (await self.tcp_pose(tcp)).to_api_model()
+        motion_group_description = await self._fetch_motion_group_description()
+        tcp_offset = (
```
biering: I guess here will be a merge conflict with the new branch

nova/cell/motion_group.py
```diff
@@ -434,6 +500,79 @@ async def _plan_with_collision_check(
...
+            )
+
+        request: api.models.PlanCollisionFreeRequest = api.models.PlanCollisionFreeRequest(
+            motion_group_setup=motion_group_setup,
+            start_joint_position=api.models.DoubleArray(start_joint_position),
+            target=api.models.DoubleArray(list(target_joint_positions)),
+            algorithm=action.algorithm,
+        )
+
+        response: api.models.PlanCollisionFreeResponse = (
```
biering: Is typing required here?

nova/utils/collision_setup.py
```diff
@@ -70,15 +70,13 @@ def motion_group_setup_from_motion_group_description(
     tool = api.models.Tool(tool_colliders.root) if tool_colliders is not None else None
     link_chain = (
         api.models.LinkChain(
-            list(
-                api.models.Link(link.root)
-                for link in motion_group_description.safety_link_colliders
-            )
+            list(api.models.Link(link) for link in motion_group_description.safety_link_colliders)
```
biering:
```suggestion
            [api.models.Link(link) for link in motion_group_description.safety_link_colliders]
```

nova/utils/collision_setup.py
```diff
@@ -70,15 +70,13 @@ def motion_group_setup_from_motion_group_description(
...
         )
         if motion_group_description.safety_link_colliders
         else None
     )
     collision_setup = api.models.CollisionSetup(
+        # Question to Dirk: does this mean safety zones are forbidden areas? Is it possible to enter but with low speed?
```
biering: I guess this needs to be answered by Markus S. but I think they are forbidden

tests/cell/test_motion_group.py
```diff
@@ -17,21 +16,22 @@ async def test_empty_list():
...
-        colliders=api.models.ColliderDictionary({collider_id: collider})
-    )
-
-
 @pytest.fixture
 def collision_setup():
-    return create_collision_setup()
+    return api.models.CollisionSetup(
+        colliders=api.models.ColliderDictionary(
+            root={
```
biering:
```suggestion
            {
```

PR #348, opened 2025-12-04
feat(NDX-43)!: release Python SDK for NOVA v2

biering:
# Python SDK for NOVA API v2 Release Notes & Migration Guide
With this release we start using v2 APIs NOVA platform from Python SDK.

## MotionSettings
* `position_zone_radius` is deprecated and replaced by `blending_radius`
* `min_blending_velocity` is deprecated and replaced by `blending_auto`

## Collision Free Motion Planning
* New parameter `algorithm` added. RRT is used as default. Check `nova.actions.motions.collision_free` for more information.
* When using collision free planning with a target pose, inverse kinematics are used to find which joint positions will be used to reach that pose. This ensures the shortest norm distance, see nova/cell/motion_group.py._find_shortest_distance, but can cause unwanted movements in the robot. In upcoming releases, this automatic behavior will be replaced by an API similar to: `https://portal.wandelbots.io/docs/api/v1/ui/#/operations/calculateInverseKinematic`.
* NOVA API V2 supports layer collisions and this is now supported by the SDK. The SDK deep copies the motion group setup while planning. Since collision scenes can be big, this can cause unexpected memory consumption. Please leave feedback with customersuccess@wandelbots.com if you'd like us to optimize the memory usage.

## Module restructuring
* Moved `motion_group.py`, `controller.py`, `io.py` and `movement_controllers.py` to `/cell` module
* Streamlined `nova.api` API access by using only `v2_pydantic` of the autogenerated API client

## Pose
Make Pose handle `None` and `api.models.Pose` in the constructor:
* `Pose(position=(1, 2, 3), orientation=None)` > `Pose(api.models.Pose(position=None, orientation=api.models.RotationVector([4, 5, 6])))`
* `Pose(position=None, orientation=(4, 5, 6))` > `Pose(position=Vector3d(x=0.0, y=0.0, z=0.0), orientation=Vector3d(x=4.0, y=5.0, z=6.0))`
* `Pose(None)` > `Pose(position=Vector3d(x=0.0, y=0.0, z=0.0), orientation=Vector3d(x=0.0, y=0.0, z=0.0))`

## Motion Group
* Added forward kinematics
* `compare_collision_scenes`, `validate_collision_scenes`, `combine_trajectories` moved to utils
* Added `get_model`
* Added `get_setup`
* Added `get_mounting`
* Added `get_safety_collision_setup`
* Added `get_default_collision_link_chain`
* Added `tcp_offset`
* `movement_controller` interface has changed as per NOVA API.
* If the robot is not at the start point of the trajectory, you will get an error and the movement will not happen. If you don't explicitly provide a start position, current joint configuration will be used as starting point.

## Others
* `Nova` class now exposes an ApiGateway with a field named `api`. `Internal_api_client` will be removed in a future release

## Breaking Changes
* `combine_trajectory` is removed from root `__init__.py`
* Changed `Pose.to_api_pose` to `Pose.to_api_model`
* Changed MotionGroup `.motion_group_id` to `.id`
* Changed Controller `.controller_id` to `.id`
* Upgraded minimum Python version to 3.11
* Removed nova.nats package. `Nova` class exposes a library integration from nova.nats field.
* Removed Auth0 package and related environment variables
* Logger changes:
  * Removed `LOGGER_NAME` environment variable. Top level logger is named “nova”. Each python sub-module logs to its own logger.
  * Removed deprecated `nova.core.logger` module
* NovaConfig:
  * Makes host in NovaConfig required
  * Remove deprecated username and password environment variables
  * Host and NOVA_API parameter requires explicit schema, http or https
* When no motion settings are provided, default motion settings with max velocity set to 50 and all else not set are used
* `Cell.nats` return nats library instance instead of NatsClient
* Removed `nova/core/env_handler`
* ApiGateway now accepts NovaConfig as constuctor parameter
* Fixed typo in `extentions` module and renamed it to `extensions`
* Removed `Controller.cell_id`
* Response type of the `motion_group` functions, `stream_execute`, `stream_plan_and_execute`, changes from `AsyncIterable` to `AsyncGenerator`

### DIFF
--- a/nova_rerun_bridge/blueprint.py
+++ b/nova_rerun_bridge/blueprint.py
@@ -18,10 +18,11 @@ def get_blueprint(motion_group_list: list[str]) -> rrb.Blueprint:
     )
 
 
-def send_blueprint(motion_group_list: list[str]) -> None:
+def send_blueprint(motion_group_list: list[str], include_coordinate_system: bool = True) -> None:
     """Send blueprint with 3D view.
 
     Args:
         motion_group_list: List of motion group names
+        include_coordinate_system: Currently unused flag to align with caller expectations.
     """
     rr.send_blueprint(get_blueprint(motion_group_list))
--- a/nova_rerun_bridge/nova_rerun_bridge.py
+++ b/nova_rerun_bridge/nova_rerun_bridge.py
@@ -92,7 +92,7 @@ class NovaRerunBridge:
         rr.reset_time()
         rr.set_time(TIME_INTERVAL_NAME, duration=0)
 
-        send_blueprint(motion_groups)
+        send_blueprint(motion_groups, True)
         self.log_coordinate_system()
 
     def log_coordinate_system(self) -> None:
--- a/nova_rerun_bridge/robot_visualizer.py
+++ b/nova_rerun_bridge/robot_visualizer.py
@@ -1,4 +1,5 @@
 import re
+from typing import Any, cast
 
 import numpy as np
 import rerun as rr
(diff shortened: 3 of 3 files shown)


nova/cell/motion_group.py
```diff
@@ -0,0 +1,737 @@
...
+            tg.create_task(execution(), name=f"execute_trajectory-{trajectory_id}-{self.id}")
+
+            while (motion_group_state := await states.get()) is not SENTINEL:
+                assert isinstance(motion_group_state, api.models.MotionGroupState)
+                yield motion_group_state_to_motion_state(motion_group_state)
+
+            # when the execution task finished
+            # task group will still wait for the monitoring task
+            # so we need to cancel it
+            monitor_task.cancel()
```
mahsumdemirwb: Check with @biering and @functionistic 
if the cancelation will propagate to the use code,
in the tests we didn't have this

lollyfy (approved):

PR #350, opened 2025-12-10
fix(NDX-318): Unclosed client session

functionistic:


nova/program/runner.py
```diff
@@ -499,6 +497,10 @@ async def _run_program(
             # Handle any exceptions raised during entering the robot cell context
             self._handle_general_exception(exc)
             await self._set_program_state(api.models.ProgramRunState.FAILED, on_state_change, nova)
+        finally:
+            # TODO not the most elegant way to close nova instance, especially since we seem to not know
+            # here if we created it or not
+            await nova.close() if nova is not None else None
```
biering: Why not


```suggestion
            if nova is not None: # or if nova:
                await nova.close()
```

PR #356, opened 2025-12-18
fix(NDX-448): correctly find the best solution in collision_free

biering:
Sometimes collision_free movements can lead to huge joint configuration changes even that the actual cartesian pose is not far away. The MR ensures that solutions are ordered and prevent that these huge joint configuration changes are not happening if possible while staying in the robot limits.

- Updated collision_free_p2p.py example:

image

### DIFF
--- a/nova/program/runner.py
+++ b/nova/program/runner.py
@@ -143,8 +143,13 @@ async def _ensure_preconditions(
     return created_controllers
 
 
-async def _cleanup_preconditions(nova: Nova, controller_ids: list[str]) -> None:
+async def _cleanup_preconditions(
+    nova: Nova, preconditions: ProgramPreconditions | None, controller_ids: list[str]
+) -> None:
     """Clean up controllers by their IDs."""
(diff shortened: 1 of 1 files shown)


examples/collision_free_p2p.py
```diff
@@ -85,10 +132,10 @@ async def collision_free_p2p(ctx: nova.ProgramContext) -> None:
     """
     nova = ctx.nova
     cell = nova.cell()
-    controller = await cell.controller("ur5")
+    controller = await cell.controller("kuka-kr16-r2010")
 
-    await ctx.nova.api.virtual_robot_setup_api.set_virtual_controller_mounting(
-        cell=cell.id,
+    await nova._api_client.virtual_robot_setup_api.set_virtual_controller_mounting(
```
mahsumdemirwb:
```suggestion
    await nova.api.virtual_robot_setup_api.set_virtual_controller_mounting(
```

mahsumdemirwb (commented):
added a comment about np.round

PR #376, opened 2026-01-23
feat: integrate bus io to sdk

mahsumdemirwb:
With this MR, we are integration bus io service to python SDK.
You can set bus io signals as part of your action array.
Also, several constructs are provided for building robot programs that interact with bus io:
- Set bus io values
- Read bus io values
- Wait for bus io values

### DIFF
--- a/tests/cell/test_io.py
+++ b/tests/cell/test_io.py
@@ -368,7 +368,9 @@ async def test_wait_io(setup_virtual_profinet: tuple[str, ...]):
 
             return False
 
-        wait_task = asyncio.create_task(wait_for_bus_io([test_bool], on_change=on_change, nova=nova))
+        wait_task = asyncio.create_task(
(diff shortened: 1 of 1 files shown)


nova/utils/io.py
```diff
@@ -0,0 +1,149 @@
...
+        elif isinstance(value, float):
+            io_value = api.models.IOFloatValue(io=io, value=value)
+        else:
+            raise ValueError(f"Invalid value type {type(value)}. Expected bool, int or float.")
+        io_value_list.append(api.models.IOValue(io_value))
+
+    await nova.api.bus_ios_api.set_bus_io_values(cell=cell, io_value=io_value_list)
+
+
+async def wait_for_io(
```
functionistic: There is `wait_for_bool_io` in `cell.io` which has the same intention as this new function, but does it with controller IOs. I'd like to see this unified or more explicitly separated.

PR #410, opened 2026-03-26
fix: zero-valued MotionSettings overrides and blending handling

mahsumdemirwb:
### Summary
- Fix `MotionSettings` checks so explicit `0` values are treated as set values instead of falsy/missing
- Preserve zero-valued limit overrides when building motion commands
- Add focused `MotionSettings` tests for zero-valued limits and blending edge cases
- Include lockfile updates in `uv.lock`

### Testing
- `uv run pytest tests/nova/types/test_motion_settings.py -q`


mahsumdemirwb:
@copilot check this MR, especially for possible side effects to the clients

PR #412, opened 2026-04-09
feat: make TCP optional for plan/execute to support motion groups without TCP

functionistic:
## Summary

Some motion groups don't have a TCP configured. Previously, the SDK required `tcp: str` as a mandatory parameter for `plan()`, `execute()`, and all related methods. This change makes TCP optional (`str | None`) so that joint-space-only workflows (e.g. `jnt()` actions) can work without specifying a TCP.

## Changes

- **`nova/cell/robot_cell.py`** — Changed `tcp: str` → `tcp: str | None = None` in `plan()`, `plan_and_execute()`, `stream_plan_and_execute()`, and `tcp: str | None` in `execute()`, `stream_execute()`, `_execute()`. Updated `_log_planning_results()` and `_log_planning_error()`.
- **`nova/cell/motion_group.py`** — Updated `_plan()`, `_execute()`, `_plan_with_collision_check()`, `_plan_collision_free()`, `_load_planned_motion()`, `_tune_trajectory()` to accept `str | None`. Added validation in `_plan()`: raises `ValueError` if cartesian actions (`lin`, `ptp`, `cir`, `collision_free` with Pose target) are used without a TCP. Kept `_inverse_kinematics()` as `tcp: str` since IK fundamentally requires a TCP.
- **`nova/viewers/base.py`**, **`nova/viewers/manager.py`**, **`nova/viewers/rerun.py`** — Updated `log_planning_success()` and `log_planning_failure()` to accept `tcp: str | None`. Rerun viewer skips trajectory visualization when `tcp is None`.

## Usage

```python
# Joint-space motions — no TCP needed:
await motion_group.plan_and_execute([jnt(joints)])

# Cartesian motions — TCP still required (clear error if missing):
await motion_group.plan_and_execute([lin(pose)], tcp="Flange")
```

## Verification

- All existing tests pass (non-breaking — existing code passes explicit TCP strings)
- mypy clean, ruff clean

### DIFF
--- a/nova/cell/motion_group.py
+++ b/nova/cell/motion_group.py
@@ -9,7 +9,7 @@ import numpy as np
 from nova import api
 from nova.actions import Action, CombinedActions, MovementController, MovementControllerContext
 from nova.actions.mock import WaitAction
-from nova.actions.motions import CollisionFreeMotion
+from nova.actions.motions import CartesianPTP, Circular, CollisionFreeMotion, Linear
 from nova.config import ENABLE_TRAJECTORY_TUNING
 from nova.core.gateway import ApiGateway
 from nova.exceptions import LoadPlanFailed, NoInverseKinematicsSolutionFound, PlanTrajectoryFailed
@@ -486,7 +486,7 @@ class MotionGroup(AbstractRobot):
         )
 
     async def _load_planned_motion(
-        self, joint_trajectory: api.models.JointTrajectory, tcp: str
+        self, joint_trajectory: api.models.JointTrajectory, tcp: str | None
     ) -> str:
         load_plan_response = await self._api_client.trajectory_caching_api.add_trajectory(
             cell=self._cell,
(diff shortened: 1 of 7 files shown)


nova/cell/motion_group.py
```diff
@@ -667,13 +668,27 @@ async def _plan_collision_free(
...
-        tcp: str,
+        tcp: str | None = None,
         start_joint_position: tuple[float, ...] | None = None,
         motion_group_setup: api.models.MotionGroupSetup | None = None,
     ) -> api.models.JointTrajectory:
         if not actions:
             raise ValueError("No actions provided")
 
+        if tcp is None:
+            _requires_tcp = (Linear, CartesianPTP, Circular)
```
Roberto-Wandelbots, line 680: So, what kind of movements will be valid without a TCP? Only joint-space actions like the error tells?

functionistic: Yes, that's how it is implemented currently. The API seems to allow this for all motions, because it just takes a `tcp_offset` which defaults to all zeros. But for a user we found this to be too little of a restriction, that's why we made the TCP mandatory in the first place.

nova/cell/motion_group.py
```diff
@@ -613,6 +613,7 @@ async def _plan_collision_free(
 
         best_joint_solutions: list[tuple[float, ...]] = []
         if isinstance(action.target, Pose):
+            assert tcp is not None, "TCP is required for collision_free with Pose target"
```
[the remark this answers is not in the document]
biering: I'm with copilot there

biering (approved):
lgtm

PR #413, opened 2026-04-16
fix(NDX-665): check initial state in wait_for_bus_io before listening for changes

functionistic:
## Problem

`wait_for_bus_io` never evaluates the initial IO state against the `on_change` callback. If the target condition is already met when the function is called, callers block indefinitely waiting for a change that already happened.

## Solution

After subscribing to NATS and fetching the current IO values, `on_change` is now called once with the initial state (`old_value=None`, `new_value=<current value>`). If the callback returns `True`, the function returns immediately without entering the message loop.

This is consistent with the existing `IOChange` docstring which defines `old_value=None` as "the IO did not exist previously."

## Changes

- **`nova/utils/io.py`**: Added initial `on_change` call after fetching `old_values`, updated docstring
- **`tests/cell/test_io.py`**: Added `test_wait_io_initial_state_already_met` integration test

## Backward compatibility

Existing callbacks that check `old_value is False` (like in the existing tests and examples) are unaffected — the initial call has `old_value=None`, which doesn't match `False`.

### DIFF
--- a/nova/utils/io.py
+++ b/nova/utils/io.py
@@ -96,7 +96,7 @@ async def wait_for_bus_io(
     Underlying NATS subscription guarantees at-most-once delivery when the consumer is healthy.
 
     Provided nats_subscription_kwargs should be used to tune the subscription behavior.
-    Additional tuning behaviour can be achieved by adjusting the NATS client configuration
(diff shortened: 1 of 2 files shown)


biering (approved):
lgtm

PR #420, opened 2026-05-18
feat(NDX-723): improve motion limits onboarding in start_here example

mxhrbrgr:
Fetches robot max velocity/acceleration limits dynamically from the controller description and demonstrates multiple MotionSettings presets (aggressive, soft-fast, soft-slow) to guide users toward sensible defaults.

### DIFF
--- a/examples/start_here.py
+++ b/examples/start_here.py
@@ -63,12 +63,9 @@ async def start(
     )
 
     # Define different motion settings based on the robot's limits
-    aggressiveFast = MotionSettings(tcp_velocity_limit=max_tcp_velocity, joint_acceleration_limits=max_joint_accelerations)
-    softFast = MotionSettings(tcp_velocity_limit=max_tcp_velocity, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
-    softSlow = MotionSettings(tcp_velocity_limit=250, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
-
-    # Set soft accelerations and slow velocity as used motion settings
-    motion_settings = softSlow
+    aggressive_fast = MotionSettings(tcp_velocity_limit=max_tcp_velocity, joint_acceleration_limits=max_joint_accelerations)
+    soft_medium = MotionSettings(tcp_velocity_limit=250, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
+    soft_slow = MotionSettings(tcp_velocity_limit=100, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
 
     home_joints = await motion_group.joints()
     tcp_names = await motion_group.tcp_names()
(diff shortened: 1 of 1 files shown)


examples/start_here.py
```diff
@@ -53,10 +53,23 @@ async def start(
...
+
+    # Retrieve the robot's maximum velocity and acceleration limits from its description
+    description = await motion_group.get_description()
+    max_tcp_velocity = description.operation_limits.auto_limits.tcp.velocity
+    max_joint_accelerations = tuple(
+        j.acceleration for j in description.operation_limits.auto_limits.joints
+    )
+
+    # Define different motion settings based on the robot's limits
+    aggressiveFast = MotionSettings(tcp_velocity_limit=max_tcp_velocity, joint_acceleration_limits=max_joint_accelerations)
```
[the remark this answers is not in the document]
mxhrbrgr: It is intentionally created and not used, so users see some examples on how to create such motion settings

examples/start_here.py
```diff
@@ -53,10 +53,23 @@ async def start(
...
+    softFast = MotionSettings(tcp_velocity_limit=max_tcp_velocity, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
```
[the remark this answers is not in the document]
mxhrbrgr: It is intentionally created and not used, so users see some examples on how to create such motion settings

examples/start_here.py
```diff
@@ -53,10 +53,23 @@ async def start(
...
+    softSlow = MotionSettings(tcp_velocity_limit=250, joint_acceleration_limits=tuple(a * 0.7 for a in max_joint_accelerations))
```
biering: we use snake_case

examples/start_here.py
```diff
@@ -53,10 +53,23 @@ async def start(
...
+
+    # Set soft accelerations and slow velocity as used motion settings
+    motion_settings = softSlow
```
biering: I wouldn't do this and pass the actual variables directly to the actions. there is no benefit to move it to another variable here (other than worse readability)

examples/start_here.py
```diff
@@ -67,22 +80,22 @@ async def start(
 
     # Actions define the sequence of movements and other actions to be executed by the robot
     actions = [
-        joint_ptp(home_joints, settings=normal),  # Move to home position slowly
-        cartesian_ptp(target_pose, settings=fast),  # Move to target pose
+        joint_ptp(home_joints, settings=motion_settings),  # Move to home position slowly
```
biering: I dont really understand the benefit of the change. now we always use one motion settings everywhere. the point was to show how the user can change the motion_settings in one trajectory

mxhrbrgr: addressed

examples/start_here.py
mxhrbrgr: addressed

examples/start_here.py
mxhrbrgr: addressed

PR #422, opened 2026-05-29
feat: add on-demand robot model downloading from NOVA API

mkhe93:
Replace the manual `download-models` script requirement with automatic on-demand model downloading when using the Rerun viewer.

Changes:
- dynamically fetch GLB models from the NOVA API's motion-group-models endpoint
- Integrate downloading in `trajectory.py` and `stream_state.py` before creating `RobotVisualizer` instances and caching the entire RobotVisualizer `_visualizer_cache: dict[str, RobotVisualizer] = {}`
- Remove the obsolete `_ensure_models_exist()` method that only printed a warning message
- add unit tests for `model_loader.py` helper and the `_visualizer_cache` functionality

Models are now downloaded automatically when needed during trajectory visualization or state streaming, eliminating the need to run `uv run download-models` before using the Rerun viewer.

Related Issue: https://wandelbots.atlassian.net/jira/software/c/projects/NDX/boards/1345?assignee=712020%3A6e4c3bef-37bb-443d-b7e3-71e0f6a460cd&selectedIssue=NDX-533


functionistic (commented):
What if the model on the API side is newer than the local one?

nova_rerun_bridge/model_loader.py
```diff
@@ -0,0 +1,28 @@
+"""Load robot models directly from the NOVA API."""
+
+from typing import TYPE_CHECKING
+
+from loguru import logger
+
+if TYPE_CHECKING:
+    from nova.core.gateway import ApiGateway
```
adozzer: Can nova rerun bridge work without nova? If not, why a type-checking flag?

mkhe93: probably not but I found that as a common pattern several times in the code and as it's only for type checking, the explanation of this pattern also seemed to suit for me :)
- https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

functionistic (approved):

PR #426, opened 2026-06-04
fix: complete trajectories on standstill when terminal event is dropped

rksr:
## Problem

`MotionGroup.execute()` can hang forever at `waiting for completion or error`.

Root cause is **upstream in `robotics/wbr`** (`RAEv2_ProtoRobotState`): the streamed `standstill` (physical) and the `execute`/`END_OF_TRAJECTORY` trajectory details (logical motion command) are independent signals, and `execute` is dropped the instant the executor is torn down at `MOTION_ENDED`. The terminal `END_OF_TRAJECTORY + standstill + execute` frame the SDK requires therefore exists only in a race window between teardown and the (lagging) standstill detector — which can be missed, leaving the SDK waiting forever for a frame that never arrives.

> Upstream issue: `<link robotics/wbr issue here>`

## Fix (client-side mitigation)

Complete on the **reliable** signal (`standstill`) rather than the transient one:

- **A — state machine** (`TrajectoryExecutionMachine`): once `TrajectoryEnded` has been seen (machine in `ending`), a bare standstill frame completes the move even when that frame has no `execute` block. Single source of truth, so `TrajectoryCursor` inherits it too.
- **B — standstill watchdog** (`stall_watchdog.py`): backstop for the case `TrajectoryEnded` is never observed at all (machine stuck in `executing`) — complete if at target, otherwise fail via a hard ceiling (`max_stall_s`) so `execute()` can never hang. Raises `MovementStalled` (subclass of `ErrorDuringMovement`, so existing handlers still catch it).

## Behavior change

A previously-hung `execute()` now either **completes** or **raises `MovementStalled`**. A standstill in `ending` away from target now **completes** (trusting the controller's explicit `TrajectoryEnded`) rather than erroring — this avoids false failures on a linear rail where a tight joint tolerance would otherwise trip.

## Validation

Reproduced the hang and verified the fix end-to-end on a live virtual `ur5e`:
swallowed terminal frame, `execute=None` standstill, never-delivered `TrajectoryEnded`, zero-motion, legitimate mid-path dwell (must not abort), and `WAITING_FOR_IO` standstill (must not trip). Unit + full `tests/cell` suites green; `ruff` + `mypy` clean.

**Not yet validated on real ABB / external-axis hardware.**

## Follow-ups (not in this PR)

- Expose `stall_timeout_s` / `max_stall_s` / `at_target_tolerance` via the public API (currently context-level defaults).
- Observability counter for watchdog/standstill-after-end completions (to measure how often the upstream race fires).
- Real-ABB + external-axis validation.
- The upstream `robotics/wbr` fix (emit/keep the terminal frame until standstill is confirmed).


biering:
was just open for reference.

PR #435, opened 2026-06-24
feat: update collision setup example

mxhrbrgr:
- don't use nova._api private field
- added example line on how to load an existing collision setup by id correctly
- removed all the rerun_bridge stuff. Not required, as all of that gets logged automatically nowadays


biering (approved):
lgtm

PR #437, opened 2026-06-29
feat: made novax usage easier e.g. autoscanning

biering:
## Make novax programs easier to register and serve

Previously each program needed an explicit `register_program(...)` call and hand-wired FastAPI setup. Now `@nova.program` functions self-register and wiring is a single line.

### Implicit registration + one-line setup
```python
import nova
from fastapi import FastAPI

@nova.program(id="hello")
async def hello(ctx: nova.ProgramContext):
    ...

app = FastAPI()
novax = nova.Novax(app)  # includes router + auto-registers all imported programs
```
- `@nova.program` adds itself to a global registry (`nova/program/registry.py`).
- `Novax(app)` includes the programs router and auto-registers; `auto_register()` / `register_module()` / `register_program()` remain available.
- `Novax` is importable from the SDK (`from nova import Novax`) when the optional `novax` extra is installed, with a clear error otherwise:
  ```
  The 'novax' package requires the optional 'novax' extra, which is not installed.
  Install it with: uv sync --extra novax
  ```

### One-command dev serve (no Docker)
```bash
uv run novax run examples/start_here.py --cell cell
```
- New `novax` CLI: `novax run <module|file> [--host] [--port] [--cell]`.
- Missing `CELL_NAME` is now an info-level dev hint instead of an error.

### Misc
- Slimmed the example app + README to the one-liner.
- Tests: `tests/novax/test_auto_register.py` (registry dedupe + `auto_register`).

### DIFF
--- a/examples/your-nova-app/your_nova_app/register_programs.py
+++ b/examples/your-nova-app/your_nova_app/register_programs.py
@@ -22,9 +22,7 @@ app = FastAPI(
 # Include the programs router and auto-register every @nova.program in this app.
 # Importing start_here above is enough — programs register themselves on import.
 # See https://github.com/wandelbotsgmbh/wandelbots-nova/blob/main/README.md#novax for more information
-novax = Novax()
-novax.include_programs_router(app)
-novax.auto_register()
+novax = Novax(app)
 
 app.add_middleware(
     CORSMiddleware,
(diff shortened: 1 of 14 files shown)


examples/your-nova-app/your_nova_app/register_programs.py
```diff
@@ -3,9 +3,10 @@
 from fastapi import FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse, HTMLResponse
-from novax import Novax
+from your_nova_app.start_here import \
+    start  # noqa: F401  (import registers the program)
```
mkhe93: are two lines required for that behaviour?

mkhe93 (approved):

PR #440, opened 2026-07-03
chore: generate motion group models from api

mkhe93:
Generate `motion_group_models.py` from api for better UX with auto-completion.

## Changes Made
- add script which fetches the existing motion group models on a given nova instance
- execute the script after the test instance has been setup


scripts/create_motion_group_models.py
```diff
@@ -0,0 +1,55 @@
...
+        "# AUTO-GENERATED by scripts/create_motion_group_models.py — do not edit manually",
+        "",
+        "from enum import StrEnum",
+        "",
+        "",
+        "class MotionGroupModel(StrEnum):",
+    ]
+    if not models:
+        lines.append("\tpass")
+        logger.warning("API returned no models — skipping file generation.")
```
functionistic: Does this mean we can release an SDK without any models in the Enum?
Should we not fail the pipeline then?

mkhe93: I raise an exception now

nova/cell/motion_group_models.py
```diff
@@ -0,0 +1,149 @@
+# AUTO-GENERATED by scripts/create_motion_group_models.py — do not edit manually
+
+from enum import StrEnum
+
+
+class MotionGroupModel(StrEnum):
```
biering: just to bring that in here: https://typing.python.org/en/latest/spec/literal.html we could also use string literals instead of an enum. To my knowledge it should still provide auto completion with the strings then. But also a enum is totally fine for me.

biering (approved):

nova/cell/motion_group_models.py
functionistic: If literals would allow autocompletion with the righthand side string (i.e. the Enum value) of the list below, I would prefer it too.
@mkhe93 can you verify this?

mkhe93: @copilot can you verify this pls? :D

biering (approved):
lgtm

PR #441, opened 2026-07-07
feat: direction constraint

hobbeshunter:
This pull request will add the direction constrained motion types (cartesian ptp, joint ptp, collision_free with constraints).

Also adds an example workflow for direction constrained motions.


nova/actions/motions.py
```diff
@@ -173,6 +182,102 @@ def cartesian_ptp(
...
+    ...     ),
+    ...     constraint=constraint,
+    ...     settings=MotionSettings(tcp_velocity_limit=30),
+    ... )
+    DirectionConstrainedCartesianPTP(metas={}, type='direction_constrained_cartesian_ptp', target=ConstrainedPose(position=Vector3d(root=[1.0, 2.0, 3.0]), orientation=0.0), settings=MotionSettings(blending_auto=None, blending_radius=None, joint_velocity_limits=None, joint_acceleration_limits=None, joint_jerk_limits=None, tcp_velocity_limit=30.0, tcp_acceleration_limit=None, tcp_jerk_limit=None, tcp_orientation_velocity_limit=None, tcp_orientation_acceleration_limit=None, tcp_orientation_jerk_limit=None, position_zone_radius=None, min_blending_velocity=None), collision_setup=None, constraint=DirectionConstraint(world=Vector3d(root=[0.0, 0.0, 1.0]), tcp=Vector3d(root=[0.0, 1.0, 0.0]), tolerance=0.05, constraint_name='DirectionConstraint'))
+
+    """
+
+    type: Literal["direction_constrained_cartesian_ptp"] = "direction_constrained_cartesian_ptp"
+    target: api.models.ConstrainedPose
```
hobbeshunter: @functionistic do you also want this type to be in the base class? Or will this break lots of stuff elsewhere?

examples/direction_constrained.py
```diff
@@ -0,0 +1,147 @@
...
+            virtual_controller(
+                name="ur10e",
+                manufacturer=api.models.Manufacturer.UNIVERSALROBOTS,
+                type="universalrobots-ur10e",
+            )
+        ],
+        cleanup_controllers=False,
+    ),
+)
+async def plan_and_execute_direction_constrained(ctx: ProgramContext):
```
functionistic, line 63: The function name would be okay albeit verbose/long. The file name should be shorter. (the `plan_and_execute` prefix does not add meaning.

examples/plan_and_execute_direction_constrained.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    first_phase_actions = [
+        joint_ptp(home_joints, settings=normal),
+        linear(projected_pose, settings=normal),  # keep tcp position fixed
+        direction_constrained_cartesian_ptp(
+            constrained_target_pose, constraint=direction_constraint_pos_z, settings=fast
+        ),
+    ]
+    first_phase_trajectory = await motion_group.plan(first_phase_actions, tcp)
+    await motion_group.execute(first_phase_trajectory, tcp, actions=first_phase_actions)
```
functionistic: `first_phase_trajectory` only exists to be stuffed into execute on the next line. That's exactly why `plan_and_execute` exists as a convenience.
same for `second_phase_trajectory`

hobbeshunter: Done in e51c59073e22c5632c3bdc0148f0a1212aaa679f

examples/direction_constrained.py
hobbeshunter: Done in e51c59073e22c5632c3bdc0148f0a1212aaa679f

nova/actions/motions.py
hobbeshunter: Added it in e51c59073e22c5632c3bdc0148f0a1212aaa679f

What do you think?

PR #447, opened 2026-07-14
fix: Update virtual env version to fix setuptools cve

nacarl:
pip audit found a cve: https://github.com/wandelbotsgmbh/wandelbots-nova/actions/runs/29327032931 

setuptools is a dependency of virtualenv


biering (approved):
lgtm

PR #453, opened 2026-07-21
feat: added docs for program usage

biering:
Entire-Checkpoint: 57fb02c13e7d


docs/programs.md
```diff
@@ -0,0 +1,272 @@
...
+
+```bash
+nova app install .
+```
+
+This builds the container image, pushes it, and registers the app (including its
+home-screen tile) on the cell. Your programs then appear in the NOVA frontend and can be
+executed from there.
+
+### Hot reload during development (Skaffold)
```
Maddi97, line 228: We could provide  the URL to Skaffold here and explicitly mention that it is an external tool.

docs/programs.md
```diff
@@ -0,0 +1,272 @@
...
+
```
Maddi97, line 229: And maybe we should also mention that the user can configure the registy that the App is pushed to in skaffold.dev

examples/your-nova-app/app/register_programs.py
```diff
@@ -13,16 +14,16 @@
...
     version="0.1.0",
     description="An application that serves your robot programs 🦾",
     root_path=BASE_PATH,
 )
 
-# Include the programs router and register every imported @nova.program. Importing
-# the program module above is enough -- the decorator self-registers on import.
-# TODO: switch to directory scanning (``Novax(app, programs_dir="programs")``) once
-# that SDK feature is released, so new programs are picked up without a manual import.
+# Include the programs router and scan the ``programs`` directory: every @nova.program
```
Maddi97:
```suggestion
# Include the programs router and scan the directory passed to Novax via the ``programs_dir``
# argument: every @nova.program
```

Maddi97 (approved):

PR #455, opened 2026-07-23
feat(NDX-810)!: integrate DatasetPose into SDK

Maddi97:
Integrates the new API model type DatasetPose into the SDK. Note: This type is not released yet and thus temporarily duplicated in the SDK type.

This PR extends nova.types.Pose to accept additional API model inputs (notably DatasetPose / ConfiguredPose)

Changes:

- Add Pose constructor dispatch (_resolve_pose) plus new Pose.from_* helpers to support API model inputs and preserve kinematic_configuration.
- Update motion factory functions (linear, cartesian_ptp, circular, spline) to convert targets via Pose(...) directly.
- Add new tests covering allowed/forbidden Pose constructor forms and motion target conversion behavior.
- Temporarily add api models as redundant types into sdk
- restricts pose matrix multiplication to pose operand only as constructor is very convenient and mult type can be more explicit

### DIFF
--- a/nova/actions/motions.py
+++ b/nova/actions/motions.py
@@ -11,17 +11,6 @@ from nova.types.pose import Pose
 PoseOrSequence = Pose | Sequence[float]
 
 
-def _convert_to_pose(target: PoseOrSequence) -> Pose:
-    if not isinstance(target, Pose):
-        t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
-
-        if len(t) != 6:
-            raise ValueError("Target must be a sequence of 6 floats")
-
-        target = Pose(t)
-    return target
-
-
 class Motion(Action, ABC):
     """Base model of a motion
 
@@ -92,8 +81,7 @@ def linear(
 
     """
     if not isinstance(target, Pose):
-        t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
-        target = Pose(t)
+        target = Pose(target)
 
     kwargs.update(utils.get_caller_metas())
 
@@ -163,7 +151,7 @@ def cartesian_ptp(
     CartesianPTP(metas={'line_number': 1}, type='cartesian_ptp', target=Pose(position=Vector3d(x=1.0, y=2.0, z=3.0), orientation=Vector3d(x=4.0, y=5.0, z=6.0), kinematic_configuration=None), settings=MotionSettings(blending_auto=None, blending_radius=None, joint_velocity_limits=None, joint_acceleration_limits=None, joint_jerk_limits=None, tcp_velocity_limit=50.0, tcp_acceleration_limit=None, tcp_jerk_limit=None, tcp_orientation_velocity_limit=None, tcp_orientation_acceleration_limit=None, tcp_orientation_jerk_limit=None, position_zone_radius=None, min_blending_velocity=None), collision_setup=None)
 
     """
-    target = _convert_to_pose(target)
+    target = target if isinstance(target, Pose) else Pose(target)
     kwargs.update(utils.get_caller_metas())
     return CartesianPTP(
         target=target, settings=settings, collision_setup=collision_setup, metas=kwargs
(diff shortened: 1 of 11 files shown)


Maddi97:
@claude Review

nova/types/pose.py
```diff
@@ -232,6 +252,8 @@ def __matmul__(self, other):
         if isinstance(other, Iterable):
             seq = tuple(other)
             return self.__matmul__(Pose(seq))
+        if isinstance(other, api.models.DatasetPose):
+            return self.__matmul__(Pose.from_dataset_pose(other))
 
         raise ValueError(f"Cannot multiply Pose with {type(other)}")
```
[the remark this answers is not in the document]
Maddi97: Fixed in 1b7f6dcbcc88631ebaa68cae9319e618a643b93e

nova/actions/motions.py
```diff
@@ -92,8 +81,7 @@ def linear(
 
     """
     if not isinstance(target, Pose):
-        t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
-        target = Pose(t)
+        target = Pose(target)
```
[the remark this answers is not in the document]
Maddi97: Good Question: Was there an intended support for different iterables such as np.arrays?

nova/actions/motions.py
```diff
@@ -91,9 +82,7 @@ def linear(
     Linear(metas={'line_number': 1}, type='linear', target=Pose(position=Vector3d(x=1.0, y=2.0, z=3.0), orientation=Vector3d(x=4.0, y=5.0, z=6.0), kinematic_configuration=None), settings=MotionSettings(blending_auto=None, blending_radius=None, joint_velocity_limits=None, joint_acceleration_limits=None, joint_jerk_limits=None, tcp_velocity_limit=50.0, tcp_acceleration_limit=None, tcp_jerk_limit=None, tcp_orientation_velocity_limit=None, tcp_orientation_acceleration_limit=None, tcp_orientation_jerk_limit=None, position_zone_radius=None, min_blending_velocity=None), collision_setup=None)
 
     """
-    if not isinstance(target, Pose):
-        t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
-        target = Pose(t)
+    target = target if isinstance(target, Pose) else Pose(target)
```
biering, line 85: Here is a breaking change with numpy:

```python
def linear(target):
    t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
    return Linear(Pose.from_tuple(t))
```

This was possible before, now not anymore:

```python
linear(np.array([100, 200, 300]))
```

I just want to raise this because A) its a breaking change (we don't know if its used somewhere in any customer project or not) and B) if we want to explicitly allow np in the Pose constructor or not?

nova/actions/motions.py
```diff
@@ -8,18 +8,9 @@
...
-def _convert_to_pose(target: PoseOrSequence) -> Pose:
-    if not isinstance(target, Pose):
-        t = (*target, 0.0, 0.0, 0.0) if len(target) == 3 else target
-
-        if len(t) != 6:
-            raise ValueError("Target must be a sequence of 6 floats")
-
-        target = Pose(t)
-    return target
+PoseCompatible = (
```
Maddi97: Question: Should I rather put this to Pose? And to avoid Breaking Changes, the Pose constructor should accept all these types as input (e.gt his would resolve @biering https://github.com/wandelbotsgmbh/wandelbots-nova/pull/455/changes#r3663818963)

Additionally, `Pose.__matmul__` 'other'-Parameter (right side factor) should probably have`PoseCompatible` as type. Right now it accepts `Iterable` (Pose.py line 254) which would include the Sequence. But Iterable would also accept all types that implement the __iter__ thunder method. E.g. this would include a String as argument which would break the multiplication

nova/types/pose.py
```diff
@@ -86,19 +101,23 @@ def __init__(self, *args, **kwargs):
         >>> kc2 = api.models.KinematicConfiguration(kinematic_branch=kb, axis_ranges=ar)
         >>> Pose((1, 2, 3, 4, 5, 6), kinematic_configuration=kc2).kinematic_configuration == kc2
         True
+        >>> Pose(api.models.DatasetPose(id='p1', pose=api.models.Pose(position=api.models.Vector3d([1, 2, 3]), orientation=api.models.RotationVector([4, 5, 6]))))
+        Pose(position=Vector3d(x=1.0, y=2.0, z=3.0), orientation=Vector3d(x=4.0, y=5.0, z=6.0), kinematic_configuration=None)
```
Maddi97, line 106: Question: Is this second example with testing the kinematic_configuration necessary?

nova/types/pose.py
```diff
@@ -197,7 +216,7 @@ def to_api_model(self) -> api.models.Pose:
             ),
         )
 
-    def __matmul__(self, other):
+    def __matmul__(self, other) -> Pose:
```
Maddi97: Thought: other should be of type `PoseCompatible` and instead accepting`Iterable` we should accept sequence[float] instead.

Related to: https://github.com/wandelbotsgmbh/wandelbots-nova/pull/455/changes#r3664024464

nova/actions/motions.py
Maddi97: We agreed on adding explicitly allowing np.array and not to introduce a breaking change.

Added in c6c31bb6ebaf5f9289aae0faa6300f4b4d5aa0ca

nova/types/pose.py
```diff
@@ -1,44 +1,60 @@
...
+        return pose
+    if len(args) == 1 and isinstance(args[0], api.models.ConfiguredPose):
+        configured_pose = args[0]
+        if (
+            configured_pose.kinematic_configuration is not None
+            and kinematic_configuration is not None
+        ):
+            raise ValueError(
+                "Cannot specify kinematic_configuration when passing a PoseLike "
+                "with a kinematic_configuration"
```
functionistic:
```suggestion
                "Cannot specify kinematic_configuration when passing a ConfiguredPose"
```

nova/types/pose.py
functionistic: `PoseLike`?

functionistic (approved):

nova/types/pose.py
Maddi97: As discussed: From now on Pose matrix multiplication is only allowed on Pose types for both operands

biering (approved):

PR #463, opened 2026-07-30
feat: add dev-wheel script to build dev wheels from CLI

biering:
- added a script to trigger a new dev wheel build on the current branch (branch needs to be pushed to remote already)

image


scripts/trigger_dev_wheel.sh
```diff
@@ -0,0 +1,111 @@
...
+fi
+
+# --- 7) SHOW THE INSTALL STRING -----------------------------------------------
+BUILT_SHA="$(gh run view "$RUN_ID" --json headSha -q '.headSha')"
+
+echo ""
+echo "------------------------------------------------------------"
+info "📦  Dev wheel built successfully. Install it with:"
+echo ""
+echo "    pip install \"wandelbots-nova @ git+https://github.com/wandelbotsgmbh/wandelbots-nova.git@${BUILT_SHA}\""
```
Maddi97: Maybe rather `uv pip install` as uv is the standard in the repo

Maddi97 (approved):

scripts/trigger_dev_wheel.sh
biering: good point

PR #471, opened 2026-08-06
feat(cursor): carry IO overlay on every start, fix false-success completion

functionistic:
Brings `TrajectoryCursor` to parity with the `move_forward` movement controller, so a later change can collapse the two. **`move_forward` is untouched here** — the actual merge is a separate, independently revertible step.

## Why

`move_forward` and `TrajectoryCursor.cntrl` implement the *same* `executeTrajectory` protocol, and the loop has now forked five ways across `main` and four in-flight branches. `cursor.forward()` from location 0 with no target *is* `move_forward`; everything else the cursor offers is a superset. Rationale and sequencing: [`docs/architecture/incoming/move-forward-to-cursor.md`](docs/architecture/incoming/move-forward-to-cursor.md).

This PR is Phases A + B of that plan.

## The IO overlay was silently dropped

A trajectory's `io_write` actions never reached the controller when executed through a cursor — `Intent.to_commands()` built `StartMovementRequest` without `set_outputs`. This is the single most valuable overlay kind, so it is fixed first.

**Verified on hardware** (virtual KUKA) that `set_outputs` has *override* semantics, which shapes the API:

| Experiment | Result |
|---|---|
| Start `[A@1.0]`, pause before 1.0, resume with `[B@2.0]` | `A=False`, `B=True` → the second start **replaced** the first list |
| Stop exactly on `A@1.0`, reset `A` externally, resume from 1.0 | `A=True` → **re-fires at the boundary** |

Consequences:

- **Every** start must carry the full list. A resume that omits it does not "keep" the overlay, it *clears* it. So `set_outputs` is a **constructor-level default**, not a per-call argument — a per-call argument is a footgun where the first resume that forgets it disables all remaining IO.
- Boundary re-firing is accepted rather than engineered around (idempotent for level-triggered outputs; worth noting for anything edge-sensitive downstream).
- `start_on_io` / `pause_on_io` get the same treatment; per-call arguments still win.

> Heads-up for `feat/path-triggers-distance-offset`: its resume sends a bare `StartMovementRequest(direction=FORWARD)`, so under override semantics that branch drops every remaining IO write on the first blocking async action.

## Two defects found while establishing parity

**The cursor reported success for movement it never commanded.** Completion was driven purely by the state stream with no correlation to what was sent, so `await cursor.forward()` could resolve to `OperationResult(final_location=3.0, error=None)` having emitted only an `InitializeMovementRequest`. A caller would then advance its program with the robot still at the start. Completion is now gated on the operation having been *commanded* — recorded immediately before the request is yielded, since gating on the acknowledgement instead loses a race against a fast state stream. An operation whose state stream ends first is now failed rather than silently abandoned.

**`pause()` could hang.** It was only ever marked commanded by its acknowledgement, so a delayed or lost ack left the future unresolved. Now gated symmetrically with start.

## Also

- `joint_trajectory` is optional; `actions=[]` means "no action metadata" rather than asserting a zero-length trajectory (this is a supported `execute()` path today).
- The state stream may be an `AsyncIterator` **or** a zero-argument factory.
- Movement errors raise `ErrorDuringMovement`, and `cntrl` unwraps a single-exception `BaseExceptionGroup` from its `TaskGroup`.
- The state monitor tees *all* states to the cursor's iterator (bounded, drop-oldest) instead of only while an operation is active — observers need states from before movement starts.
- `emit_motion_events` suppresses the 5 Hz signal for non-interactive callers.

## A deliberate non-change

A stop still **discards** a queued intent. Flushing it was implemented and reverted: it let an explicit `cursor.forward(); cursor.detach()` cancel the caller's future *and then still command the robot to move*. A stop must win. The symptom that motivated the flush is now surfaced properly by the completion gate, as an error rather than a phantom success.

## Verification

- **365 unit tests** pass, including 14 new regression tests pinning the overlay contract and both defects
- **14 integration tests** pass against a live NOVA cell, including the existing `move_forward` paths that must not regress
- A new integration test proves the IO overlay **actually fires on hardware** through the cursor
- `ruff` and `ty` clean on the changed files

## Review notes

- The mocked state-stream fixtures emit an idle state immediately and withhold trajectory progress until the start has been observed. Streams that complete instantly tear the cursor down before the request loop gets a turn — an artefact, not the behaviour under test.
- Still open, deliberately: whether `execute()` should emit motion events by default, and whether client-side overlays (markers, async actions, guards) move to a layer *above* the cursor — argued in §3 of the plan doc.

### DIFF
--- a/docs/architecture/incoming/move-forward-to-cursor.md
+++ b/docs/architecture/incoming/move-forward-to-cursor.md
@@ -1,0 +1,622 @@
+# `move_forward` → `TrajectoryCursor` — transition plan
+
+Status: **Phases A and B implemented; C–E open.** This document argues that the two
+movement controllers should become **one** implementation, and sequences the transition.
+
+It is a companion to the `command-routine.md` analysis (an unpublished working document on
+the incoming NOVA CommandRoutine format), which flagged the missing `set_outputs` on the
+cursor as a P0 blocker; references to its sections are kept for whoever holds that draft.
+This document stands on its own for everything needed to continue the work here.
+
+Date: 2026-08-04 (implementation notes added 2026-08-05)
+SDK baseline analysed: `main` @ `3e9e513`
+Branches analysed: `feat/NDX-155-simple-custom-actions` @ `8112b15`,
+`feat/path-triggers-distance-offset` @ `5560279`, `feat/motion-guards` @ `878be25`,
+`fix/trajectory-completion-standstill-watchdog` @ `9e7e4f4`
+
+---
+
+## 0. Why now — the protocol loop has forked five ways
+
+`move_forward` and `TrajectoryCursor.cntrl` are both async generators implementing the
+*same* `executeTrajectory` websocket protocol: initialize → start state monitor →
+`StartMovementRequest` → consume responses → terminate on standstill. Every feature that
+needs to *interrupt* that loop has forked it rather than shared it:
+
+| Where | Shape | What it added |
+|---|---|---|
+| `main` `move_forward.py` (172 ln) | linear generator | baseline: init, one Start, monitor, error consumer |
+| `main` `trajectory_cursor.py` (1314 ln) | intent slot + command loop | forward/backward/pause/target, playback speed, events |
+| `feat/path-triggers-distance-offset` `move_forward.py` (~260 ln) | **+ command queue, + response consumer, + pause/resume callbacks** | async actions; re-invents the cursor's `_request_loop` and `_response_consumer` |
+| `feat/motion-guards` `guarded_move_forward.py` (385 ln) | third copy | guard evaluation + stop/resume |
+| `fix/trajectory-completion-standstill-watchdog` `move_forward.py` (+82 ln) + `stall_watchdog.py` (273 ln) | fourth copy | completes trajectories when the terminal event is dropped — **applied to `move_forward` only; the cursor still has the bug** |
+
+Two observations make the merge urgent rather than merely tidy:
+
+1. **`feat/path-triggers-distance-offset` rewrote `move_forward` into a worse cursor.** It
+   adds an `asyncio.Queue` of commands, a `PauseMovementRequest`/`StartMovementRequest`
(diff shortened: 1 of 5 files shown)


functionistic:
@claude review

functionistic:
Both review comments addressed in b3ead25.

**Error message lost on the operation future** — confirmed by reproduction before fixing. With a `MovementErrorResponse(message="E-STOP triggered")`:

```
before:  cntrl raised  : ErrorDuringMovement | ... E-STOP triggered
         future raised : ErrorDuringMovement | Motion group state stream ended before the movement completed

after:   cntrl raised  : ErrorDuringMovement | ... E-STOP triggered
         future raised : ErrorDuringMovement | ... E-STOP triggered
```

The analysis was exactly right, including the mechanism: raising cancels the state monitor via the `TaskGroup`, and its `finally` completed the operation first with the generic error. Fixed as suggested — the response consumer now fails the operation with the controller's message *before* raising, so `complete()`'s already-done guard makes the monitor's completion a no-op. This is a race my own `finally` block introduced, and it also resolves the standing `TODO` about whether a movement error should fail the operation.

Added `test_operation_future_carries_the_controller_message` as regression cover, since the existing test only asserted on the exception from iterating `cntrl()` and so never exercised this divergence.

**Empty `except`** — added the explanatory comment; behaviour unchanged. Both `TimeoutError` and `StopAsyncIteration` are acceptable there and neither is the assertion under test; what matters is that nothing was yielded, which the following assertion checks.

366 unit tests pass. Local integration re-run was inconclusive — the NOVA instance I was using has since been torn down — so CI's provisioned instance is the check that matters here.

functionistic:
Superseded by #472. This PR was auto-closed by GitHub when the branch was renamed `agents/merge-move-forward-into-cursor` → `feat/trajectory-cursor-io-overlay-parity` to match repo conventions; the rename deleted the head branch rather than retargeting, and it can't be reopened without the original branch.

Same commits, nothing lost. Both review comments here were addressed in `b3ead25` and remain readable above.

PR #472, opened 2026-08-06
feat(cursor): carry IO overlay on every start, fix false-success completion

functionistic:
> Supersedes #471, which GitHub auto-closed when this branch was renamed to match repo conventions. Same commits; the review thread on #471 (both comments addressed in `b3ead25`) is still readable there.

Brings `TrajectoryCursor` to parity with the `move_forward` movement controller, so a later change can collapse the two. **`move_forward` is untouched here** — the actual merge is a separate, independently revertible step.

## Why

`move_forward` and `TrajectoryCursor.cntrl` implement the *same* `executeTrajectory` protocol, and the loop has now forked five ways across `main` and four in-flight branches. `cursor.forward()` from location 0 with no target *is* `move_forward`; everything else the cursor offers is a superset. Rationale and sequencing: [`docs/architecture/incoming/move-forward-to-cursor.md`](docs/architecture/incoming/move-forward-to-cursor.md).

This PR is Phases A + B of that plan.

## The IO overlay was silently dropped

A trajectory's `io_write` actions never reached the controller when executed through a cursor — `Intent.to_commands()` built `StartMovementRequest` without `set_outputs`. This is the single most valuable overlay kind, so it is fixed first.

**Verified on hardware** (virtual KUKA) that `set_outputs` has *override* semantics, which shapes the API:

| Experiment | Result |
|---|---|
| Start `[A@1.0]`, pause before 1.0, resume with `[B@2.0]` | `A=False`, `B=True` → the second start **replaced** the first list |
| Stop exactly on `A@1.0`, reset `A` externally, resume from 1.0 | `A=True` → **re-fires at the boundary** |

Consequences:

- **Every** start must carry the full list. A resume that omits it does not "keep" the overlay, it *clears* it. So `set_outputs` is a **constructor-level default**, not a per-call argument — a per-call argument is a footgun where the first resume that forgets it disables all remaining IO.
- Boundary re-firing is accepted rather than engineered around (idempotent for level-triggered outputs; worth noting for anything edge-sensitive downstream).
- `start_on_io` / `pause_on_io` get the same treatment; per-call arguments still win.

> Heads-up for `feat/path-triggers-distance-offset`: its resume sends a bare `StartMovementRequest(direction=FORWARD)`, so under override semantics that branch drops every remaining IO write on the first blocking async action.

## Two defects found while establishing parity

**The cursor reported success for movement it never commanded.** Completion was driven purely by the state stream with no correlation to what was sent, so `await cursor.forward()` could resolve to `OperationResult(final_location=3.0, error=None)` having emitted only an `InitializeMovementRequest`. A caller would then advance its program with the robot still at the start. Completion is now gated on the operation having been *commanded* — recorded immediately before the request is yielded, since gating on the acknowledgement instead loses a race against a fast state stream. An operation whose state stream ends first is now failed rather than silently abandoned.

**`pause()` could hang.** It was only ever marked commanded by its acknowledgement, so a delayed or lost ack left the future unresolved. Now gated symmetrically with start.

## Also

- `joint_trajectory` is optional; `actions=[]` means "no action metadata" rather than asserting a zero-length trajectory (this is a supported `execute()` path today).
- The state stream may be an `AsyncIterator` **or** a zero-argument factory.
- Movement errors raise `ErrorDuringMovement`, and `cntrl` unwraps a single-exception `BaseExceptionGroup` from its `TaskGroup`.
- The state monitor tees *all* states to the cursor's iterator (bounded, drop-oldest) instead of only while an operation is active — observers need states from before movement starts.
- `emit_motion_events` suppresses the 5 Hz signal for non-interactive callers.

## A deliberate non-change

A stop still **discards** a queued intent. Flushing it was implemented and reverted: it let an explicit `cursor.forward(); cursor.detach()` cancel the caller's future *and then still command the robot to move*. A stop must win. The symptom that motivated the flush is now surfaced properly by the completion gate, as an error rather than a phantom success.

## Verification

- **365 unit tests** pass, including 14 new regression tests pinning the overlay contract and both defects
- **14 integration tests** pass against a live NOVA cell, including the existing `move_forward` paths that must not regress
- A new integration test proves the IO overlay **actually fires on hardware** through the cursor
- `ruff` and `ty` clean on the changed files

## Review notes

- The mocked state-stream fixtures emit an idle state immediately and withhold trajectory progress until the start has been observed. Streams that complete instantly tear the cursor down before the request loop gets a turn — an artefact, not the behaviour under test.
- Still open, deliberately: whether `execute()` should emit motion events by default, and whether client-side overlays (markers, async actions, guards) move to a layer *above* the cursor — argued in §3 of the plan doc.

---

**Note:** this branch also carries `eabe630 "deps update"` (a `uv.lock` refresh pushed by @dirksonnemann), which is unrelated to the cursor changes.

### DIFF
--- a/docs/architecture/incoming/move-forward-to-cursor.md
+++ b/docs/architecture/incoming/move-forward-to-cursor.md
@@ -183,7 +183,7 @@ Measured against `main` @ `3e9e513`, with `detach_on_standstill=True`, a cursor 
 ```python
 await self._intent_event.wait()
 if self._stop_event.is_set():
-    break                     # ← a pending intent is thrown away here
+    break  # ← a pending intent is thrown away here
 intent = self._pending_intent
 ```
 
--- a/docs/architecture/incoming/move-forward-to-cursor.md
+++ b/docs/architecture/incoming/move-forward-to-cursor.md
@@ -371,7 +371,9 @@ the overlay layer is the missing piece between the two.
 
 Each phase is independently landable and independently verifiable.
 
-> **Status: Phases A and B are implemented** on `agents/merge-move-forward-into-cursor`
+> **Status: Phases A and B are implemented** in
+> [PR #472](https://github.com/wandelbotsgmbh/wandelbots-nova/pull/472), branch
+> `feat/trajectory-cursor-io-overlay-parity`
 > (`nova/cell/movement_controller/trajectory_cursor.py`, plus
 > `tests/cell/test_trajectory_cursor_parity.py` and one new integration test).
 > Verified: 363 unit tests and 14 integration tests green against a live cell, `ruff` and
--- a/docs/architecture/incoming/move-forward-to-cursor.md
+++ b/docs/architecture/incoming/move-forward-to-cursor.md
@@ -1,0 +1,622 @@
+# `move_forward` → `TrajectoryCursor` — transition plan
+
+Status: **Phases A and B implemented; C–E open.** This document argues that the two
+movement controllers should become **one** implementation, and sequences the transition.
+
+It is a companion to the `command-routine.md` analysis (an unpublished working document on
+the incoming NOVA CommandRoutine format), which flagged the missing `set_outputs` on the
+cursor as a P0 blocker; references to its sections are kept for whoever holds that draft.
+This document stands on its own for everything needed to continue the work here.
+
+Date: 2026-08-04 (implementation notes added 2026-08-05)
+SDK baseline analysed: `main` @ `3e9e513`
+Branches analysed: `feat/NDX-155-simple-custom-actions` @ `8112b15`,
+`feat/path-triggers-distance-offset` @ `5560279`, `feat/motion-guards` @ `878be25`,
+`fix/trajectory-completion-standstill-watchdog` @ `9e7e4f4`
+
+---
+
+## 0. Why now — the protocol loop has forked five ways
+
+`move_forward` and `TrajectoryCursor.cntrl` are both async generators implementing the
+*same* `executeTrajectory` websocket protocol: initialize → start state monitor →
+`StartMovementRequest` → consume responses → terminate on standstill. Every feature that
+needs to *interrupt* that loop has forked it rather than shared it:
+
+| Where | Shape | What it added |
+|---|---|---|
+| `main` `move_forward.py` (172 ln) | linear generator | baseline: init, one Start, monitor, error consumer |
+| `main` `trajectory_cursor.py` (1314 ln) | intent slot + command loop | forward/backward/pause/target, playback speed, events |
+| `feat/path-triggers-distance-offset` `move_forward.py` (~260 ln) | **+ command queue, + response consumer, + pause/resume callbacks** | async actions; re-invents the cursor's `_request_loop` and `_response_consumer` |
+| `feat/motion-guards` `guarded_move_forward.py` (385 ln) | third copy | guard evaluation + stop/resume |
+| `fix/trajectory-completion-standstill-watchdog` `move_forward.py` (+82 ln) + `stall_watchdog.py` (273 ln) | fourth copy | completes trajectories when the terminal event is dropped — **applied to `move_forward` only; the cursor still has the bug** |
+
+Two observations make the merge urgent rather than merely tidy:
+
+1. **`feat/path-triggers-distance-offset` rewrote `move_forward` into a worse cursor.** It
+   adds an `asyncio.Queue` of commands, a `PauseMovementRequest`/`StartMovementRequest`
+   pair driven by executor callbacks, and a response consumer that tolerates pause acks.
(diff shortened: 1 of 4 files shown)


functionistic:
Two commits on this branch are **not** part of the cursor work, flagged so review can skip or split them:

- `eabe630` **deps update** (pushed by @dirksonnemann) — a `uv.lock` refresh. Welcome side effect: it clears the `uv-audit` failure that was red on #471.
- `64d8258` **ruff 0.16.1 fallout** (mine) — that refresh bumped ruff 0.15.17 → 0.16.1, which now formats Python code blocks inside Markdown and enforces `PLR0917`. CI runs `ruff format --check`, `ruff check --select I` and `ruff check`, so the bump can't go green without it. Cosmetic reformat of snippets in 10 Markdown files (comment alignment, list joining — no documentation content changed) plus three inline `# noqa: PLR0917` in `novapolicy/tests/unit/test_pose.py`, matching the existing `PLR2004` suppression in `novapolicy/gr00t/rtc.py`.

Happy to move both to their own PR if you'd rather keep this one to the cursor change — say the word and I'll rebase them out.

The cursor change itself is unchanged and still confined to `nova/cell/movement_controller/trajectory_cursor.py` plus its tests.

functionistic:
Correction to my note above: I've **split the dependency work out into #473**, so this PR is back to just the two cursor commits.

The refresh bumped ruff 0.15.17 → 0.16.1 and ty 0.0.53 → 0.0.69, which kept pulling unrelated work in here — 10 reformatted Markdown files, then 7 type diagnostics in `novapolicy`/`nova_rerun_bridge`. That belongs in its own PR, and #473 carries it (it also clears the long-standing `uv-audit` failure, and surfaced a real bug where `virtual_controller(controller_config_json=...)` was silently discarded).

**Caveat:** #473 will make this PR red once it merges — ruff 0.16.1 wants a one-line reformat of a Python snippet in `docs/architecture/incoming/move-forward-to-cursor.md`, which doesn't exist on `main` so it couldn't travel with #473. I'll rebase and fix that as soon as #473 lands, or sooner if you'd rather merge this one first.

functionistic:
Handled pre-emptively in `ad3eb2c` — the snippet is now formatted the way ruff 0.16.1 wants, which is a no-op for 0.15.17 (it doesn't format Markdown at all). Verified clean under both, so this PR stays green whichever of the two merges first. No follow-up needed.

functionistic:
Rebased onto `main` now that #473 has merged — **all 20 checks green**, including `uv-audit`, which was the last red one and is fixed by the dependency refresh that #473 carried.

The rebase was clean, no conflicts: pre-formatting the plan-doc snippet in `bbd0b38` meant the ruff 0.16.1 upgrade landed without touching this branch. Verified locally against the new toolchain (ruff 0.16.1, ty 0.0.69) before pushing — `ruff format --check`, `ruff check --select I`, `ruff check` and `ty check` all clean.

Still the same three cursor commits; nothing changed but the base. Ready for review whenever — it needs one approving review to merge.

Maddi97:
@claude review

Maddi97 (approved):

biering (approved):

functionistic:
All three review comments addressed in `fcd411d9`.

**🔴 Stop not re-checked between commands of one intent — confirmed and fixed.** The analysis was right, including the mechanism. Reproduced before touching anything, with `forward(playback_speed_in_percent=50)` and a detach after the first command:

```
intent commands : ['PlaybackSpeedRequest', 'StartMovementRequest']
future cancelled: True
before → commands sent: ['PlaybackSpeedRequest', 'StartMovementRequest']   # robot commanded after abort
after  → commands sent: ['PlaybackSpeedRequest']                           # movement withheld
```

`_request_loop` now re-checks `_stop_event` at the top of the inner loop and returns. This was a real hole in the invariant the branch argues for — a stop won over a *queued* intent but not over a *partially dispatched* one.

Added `test_stop_between_commands_of_one_intent_withholds_movement`, and verified it's genuine coverage by reverting the fix: it fails without it, passes with it. The existing test only covered single-command intents, which is exactly why this slipped through.

**🟡 `_complete_operation` duplication.** Both new failure-completion sites now route through the helper, so `final_location=self._current_location` lives in one place again.

**🟡 Test docstring and class name asserting the opposite.** Correct, and worth fixing given the hazard: the module bullet now reads that a pending intent must be *dropped* on stop and then fail rather than report a phantom success, and `TestPendingIntentSurvivesStop` → `TestStopWinsOverPendingIntent`. Assertions unchanged; they were already right.

---

On the code-quality bot's "Statement has no effect" at `test_trajectory_cursor_parity.py:504` — that's a false positive. `await future` inside `pytest.raises(...)` *is* the assertion: awaiting the future is what raises the exception being checked. Left as is.

Verified: 459 unit tests pass; `ruff format --check`, `ruff check --select I`, `ruff check` and `ty check` all clean.

PR #473, opened 2026-08-06
chore: refresh dependencies and fix ruff 0.16.1 / ty 0.0.69 fallout

functionistic:
Splits the dependency refresh out of #472 so the cursor change stays reviewable, and carries the toolchain fallout that has to land with it.

`eabe630`/`359cf03` **deps update** is @dirksonnemann's commit, originally pushed onto the cursor branch; the rest is the work needed to make it green.

## Why these have to land together

The refresh bumps two toolchain packages, and CI runs `ruff format --check`, `ruff check --select I`, `ruff check` and `ty check`:

| | before | after | consequence |
|---|---|---|---|
| ruff | 0.15.17 | 0.16.1 | now formats Python code blocks **inside Markdown**, and enforces `PLR0917` |
| ty | 0.0.53 | 0.0.69 | adds `pydantic-discarded-extra-argument` |

So the bump cannot merge on its own without turning `main` red.

## Bonus: it clears `uv-audit`

`uv-audit` has been failing on every branch (including `main`-based dependabot ones) on `aiohttp` / `cryptography` advisories. This refresh fixes it — the check is green here.

## One real bug, surfaced by the new `ty`

`api.models.VirtualController`'s field is `json_` with **alias `json`**, and the model does not set `populate_by_name`. So `json_=` was silently discarded and **`virtual_controller(controller_config_json=...)` never reached the API**.

Verified against the installed client:

```python
VirtualController(..., json_="{...}").json_   # -> None      (silently dropped)
VirtualController(..., json ="{...}").json_   # -> '{...}'   (correct)
```

Fixed by passing the alias. The pre-existing `# ty: ignore[unknown-argument]` on that line was masking it and is no longer needed.

## Everything else is cosmetic or a no-op

- **Markdown reformat, 10 files** — comment alignment and list joining inside Python snippets. No documentation content changed.
- **`PLR0917` × 3** in `novapolicy/tests/unit/test_pose.py` — inline `# noqa`, matching the existing `PLR2004` suppression in `novapolicy/gr00t/rtc.py`. These are hypothesis-driven signatures where the positional count is inherent.
- **Discarded kwargs removed** — `Sphere(position=...)` and `JointTrajectory(tcp=...)` (×4). Neither model has the field, so the values were already being thrown away at runtime; removing them changes nothing.

## Verification

`ruff format --check`, `ruff check --select I`, `ruff check` and `ty check` all clean, and the full unit suite passes (748 passed, 6 skipped) with all extras installed as CI does.

## Review note

The one judgement call worth a second opinion is the `json`/`json_` fix — it changes real behaviour, in that `controller_config_json` now actually reaches the API where before it was dropped. Anyone who set it and worked around the value being ignored would see a change.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -28,13 +28,10 @@ You can use the SDK with different Nova environments such as the Wandelbots Port
 from nova.config import NovaConfig
 
 portal_config = NovaConfig(
-    host="https://xxxxx.instance.wandelbots.io",
-    access_token="your_access_token",
+    host="https://xxxxx.instance.wandelbots.io", access_token="your_access_token"
 )
 
-ipc_config = NovaConfig(
-    host="http://192.168.0.10",
-)
+ipc_config = NovaConfig(host="http://192.168.0.10")
 ```
 
 Pass the configuration into `Nova` when you create it:
--- a/docs/programs.md
+++ b/docs/programs.md
@@ -64,8 +64,8 @@ from nova.types import Pose
 
 
 @nova.program(
-    id="hello_robot",              # unique id; defaults to the function name
-    name="Hello Robot",            # human-readable name shown in the UI
+    id="hello_robot",  # unique id; defaults to the function name
+    name="Hello Robot",  # human-readable name shown in the UI
     preconditions=nova.ProgramPreconditions(
         controllers=[
             virtual_controller(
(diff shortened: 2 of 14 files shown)


biering (approved):
lgtm

PR #483, opened 2026-08-13
fix: Prevent infinite loop when tracing hull outlines

adozzer:
## Problem

`HullVisualizer.merge_coplanar_triangles_to_polygon()` can hang forever.

Convex hulls computed from CAD collision geometry often contain coplanar faces whose boundary is **not a simple cycle** — a pinch vertex with more than two boundary neighbours (e.g. triangles that fan out from one vertex and only touch there). The boundary walk only skipped the previous vertex:

```python
for n in neighbors:
    if n != prev:
        next_vertex = n
        break
```

At a pinch vertex this can step into a sub-loop that does not contain `start`, so `next_vertex == start` never becomes true and `next_vertex is None` never happens. The `while True` loop spins forever while `loop` grows without bound.

Hit this while logging a real collision setup (681 convex hull colliders) with `nova_rerun_bridge`: `log_colliders_once()` froze on one collider at 0% CPU with steadily growing memory, so nothing was ever visualized.

## Fix

Traverse every boundary edge at most once. This bounds the walk by the number of boundary edges and terminates on non-simple boundaries, while behaviour on well-formed (simple-cycle) boundaries is unchanged.

## Test

`tests/nova_rerun_bridge/test_hull_visualizer.py` contains the minimized real-world face (8 points, 5 triangles) that triggered the hang, plus a sanity check that hulls with duplicated input vertices still yield outlines.

Verified the regression test reproduces the bug — reverting the fix makes it hit the `pytest-timeout` limit inside the `while True` loop:

```
File "tests/nova_rerun_bridge/test_hull_visualizer.py", line 33, in test_merge_coplanar_triangles_terminates_on_pinched_boundary
  polygon = HullVisualizer.merge_coplanar_triangles_to_polygon(
File "nova_rerun_bridge/hull_visualizer.py", line 101, in merge_coplanar_triangles_to_polygon
  while True:
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
```

With the fix, `tests/nova_rerun_bridge` passes (32 tests) and the 681-collider scene logs completely in ~4 s.

### DIFF
--- a/nova_rerun_bridge/hull_visualizer.py
+++ b/nova_rerun_bridge/hull_visualizer.py
@@ -93,20 +93,27 @@ class HullVisualizer:
             adj.setdefault(a, []).append(b)
             adj.setdefault(b, []).append(a)
 
-        # Walk along the boundary edges to form a closed loop
+        # Walk along the boundary edges to form a closed loop. Degenerate hulls can have
+        # pinch vertices (degree > 2) where the boundary is not a simple cycle, so every
+        # edge may only be traversed once - otherwise the walk loops forever.
         start = boundary_edges[0][0]
         loop = [start]
         current = start
         prev = None
+        visited_edges: set[tuple[int, int]] = set()
         while True:
             neighbors = adj[current]
             next_vertex = None
             for n in neighbors:
-                if n != prev:
-                    next_vertex = n
-                    break
+                if n == prev:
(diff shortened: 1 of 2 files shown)


biering (approved):
lgtm

PR #484, opened 2026-08-13
fix(nova-rerun-bridge): Derive hull outlines from trimesh face adjacency

adozzer:
Follow-up to #483, now that it is merged.

## Why

#483 stopped `merge_coplanar_triangles_to_polygon()` from looping forever, but the hand-rolled geometry that made the hang possible is still there:

1. group hull triangles into planes with a rounded normal/offset hash (`group_coplanar_triangles`)
2. walk each plane's boundary edges into a closed polygon (`merge_coplanar_triangles_to_polygon`)

Step 2 is the one that hung. Step 1 has its own problem: two triangles of the same flat face whose normals round to different keys end up in different plane groups, so the triangulation seam between them is drawn as if it were a hull edge.

## What

An edge is visible exactly when its two adjacent triangles are not coplanar — `trimesh` answers that directly:

```python
mesh = trimesh.Trimesh(vertices=points, faces=faces, process=False)
sharp = mesh.face_adjacency_angles > COPLANAR_ANGLE_TOLERANCE
return list(points[mesh.face_adjacency_edges[sharp]])
```

No plane grouping, no boundary ordering, so the class of bug fixed in #483 can no longer occur. Edges come back as two-point strips; all 12 call sites already pass the result straight to `rr.LineStrips3D` (and sometimes `compute_hull_mesh`), both of which render segments and closed polylines identically.

One subtlety worth reviewing: qhull does **not** wind hull triangles consistently, so the triangles are re-wound against the outward plane normals qhull reports in `hull.equations`. Without that, two coplanar triangles appear to meet at 180° and their shared edge survives the filter. Passing `face_normals=` to `Trimesh` does not help — trimesh recomputes them from the winding.

Deletes `plane_from_triangle`, `group_coplanar_triangles` and `merge_coplanar_triangles_to_polygon` (−158/+114 lines), including the walk guard added in #483. The planar-input fallback (`_is_coplanar` / `_compute_2d_hull_polygon`) is unchanged.

## Verification

Compared old vs new outlines over a real collision setup with 681 convex hull colliders, edge set by edge set:

| | |
|---|---|
| colliders with identical outlines | 446 / 681 |
| colliders where the new code **adds** an edge | 0 |
| colliders where the old code drew extra edges | 234 (983 edges) |
| dihedral angle of those 983 edges | ≤ 6e-5° — triangulation seams inside flat faces |
| degenerate zero-length segments emitted | 7037 → 0 (old code appended the start vertex twice) |
| time to compute all outlines | 0.27 s → 0.11 s |

So nothing visible is lost and the near-coplanar seams are gone. The remaining differences all sit within one explicit tolerance constant instead of being decided by two rounding steps in a hash key.

The topology-level regression test from #483 goes away with the function it covered; the new tests cover the contract instead: a cube outlines to exactly its 12 edges, duplicated input vertices do not change that, points lying inside a face are absorbed (the coplanar-face case that used to hang), planar input still falls back to a 2D outline, and the outline still hulls back into the same solid via `compute_hull_mesh`.

### DIFF
--- a/nova_rerun_bridge/hull_visualizer.py
+++ b/nova_rerun_bridge/hull_visualizer.py
@@ -5,6 +5,10 @@ import trimesh
 from scipy.spatial import ConvexHull
 from scipy.spatial.qhull import QhullError
 
+# Adjacent hull triangles are treated as one flat face below this angle, so the edge between
+# them is not part of the wireframe.
+COPLANAR_ANGLE_TOLERANCE = 1e-6
+
 
 class HullVisualizer:
     @staticmethod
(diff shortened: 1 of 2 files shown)


biering (approved):
lgtm

PR #487, opened 2026-08-19
fix(cursor): level-correct completion detection; guards for persistent execute state (wbr!2262)

functionistic:
## Problem

`execute()` can hang forever: trajectory completion is concluded from the `MotionGroupState` stream, where current controllers publish the terminal `TrajectoryEnded` state for a single controller step and drop the `execute` block the instant the robot settles. The shipped `TrajectoryExecutionMachine` additionally discarded every frame without an `execute` block on a wrong assumption — so even a *caught* terminal edge could strand the machine in `ending` when the confirming standstill frames arrived bare (measured on real 50 ms captures; committed as replay fixtures).

The same one-tick problem exists for pauses (pinned server-side by wbr!2322: the pause-at-standstill window is 1–2 control cycles).

Backend MR **robotics/wbr!2262** converts the published execute state to level semantics: `execute.details` persists from initialize until stop/teardown, `END_OF_TRAJECTORY` and `PAUSED_BY_USER` are re-published every step — including `PAUSED_BY_USER` **between initialize and the actual motion start**, which today's cursor would mistake for a completed pause (a *new* hang).

## Changes

1. **`TrajectoryExecutionMachine`**: a bare standstill frame now completes `ending → ended` and `pausing → paused` — after the terminal discriminator was observed, bare standstill can be the only completion signal current controllers ever deliver. No transition is fabricated from `executing`.
2. **`TrajectoryCursor` guards for level-based publishing**: `set_running()` requires motion evidence (`standstill` false or a `TrajectoryRunning` detail) instead of the mere presence of an `execute` block, and a paused state only concludes a PAUSE operation or an operation that was seen running — pre-start `PAUSED_BY_USER` frames can no longer resolve a movement operation that never moved.
3. **Tests**: synthetic streams of both wire shapes through the cursor protocol, plus committed fixtures replaying **real captures** (a 50 ms stream that previously hung in `ending`; a step-rate pause window, full and thinned to what a throttled stream delivers).
4. **Docs**: `nova/cell/movement_controller/README.md` (+ the state-machine docstring) now documents how execution state is published and consumed — the level-based semantics of wbr!2262, the legacy drop-at-settle behaviour, and the cursor's completion rules.

Deliberately **not** included (scope decision): a client-side standstill watchdog — wbr!2262 is the intended remedy; on pre-!2262 backends a never-delivered terminal edge can still hang, accepted until the backend change ships.

## Verification

- Unit suite green; ruff/ty clean.
- Live, current backend (virtual KR270): 5/5 sdk-mode runs, `execute()` returned every run; pause behaviour unchanged.
- Live, **wbr!2262 backend** (RAE pod on the dev instance switched to the MR branch): persistent `END_OF_TRAJECTORY` observed (30 frames on a 200 ms stream, previously zero → completion detected at any rate); the pre-start `PAUSED_BY_USER` window observed (2 step-rate frames) and correctly ignored — `execute()` returned at 6.3–6.5 s for a 6.01 s motion, 3/3 runs; commanded pause published persistently (501 step-rate frames).

Measurements were produced with a local capture/replay harness (state-stream observers at configurable rates, FSM replay, detector comparison); the harness and the underlying research notes are deliberately not part of this PR.

https://claude.ai/code/session_014EW8FBBuRjwttyUer2v4Lv


functionistic:
@claude review

functionistic:
Review comments addressed in 3072a8eb: the two bare `await consumer` statements flagged by github-code-quality are now `await asyncio.gather(consumer)` (identical behaviour, explicit coordination). Today's Claude review found no issues.

functionistic:
Correction to my earlier comment: 3072a8eb is reverted in 93e4cc9b. The github-code-quality "statement has no effect" findings on `await consumer` are false positives — awaiting a Task is the effect (suspension until completion + exception propagation), and the bare form is the idiomatic asyncio spelling. We keep it and dismiss the findings rather than route a single task through `asyncio.gather`.

nova/cell/movement_controller/trajectory_state_machine.py
```diff
@@ -200,14 +209,27 @@ def process_motion_state(self, state: api.models.MotionGroupState) -> StateUpdat
...
+                elif self.current_state == self.pausing:
+                    self._pause_after_standstill()
+            current_id = self._active_configuration_id()
             return StateUpdate(
                 has_execute=False,
-                state_changed=False,
+                state_changed=current_id != previous_state_id,
                 previous_state_id=previous_state_id,
-                current_state_id=previous_state_id,
+                current_state_id=current_id,
```
biering, line 232: interesting change. this will change the trajectory cursors current state we had so far, or?

biering (approved):

Maddi97 (approved):

nova/cell/movement_controller/trajectory_state_machine.py
functionistic: Yes, deliberately — but only in one direction. The behavior changes solely when the machine is already in `ending`/`pausing` (the terminal discriminator was already seen) and the confirming standstill arrives on a frame without an `execute` block. Before, that frame was discarded and the cursor hung in `ending`/`pausing` forever; now it completes to `ended`/`paused`.

That wire shape is the normal case on pre-!2262 controllers, which drop the `execute` block the instant the robot settles — the two committed fixtures are real captures of it (a 50 ms stream and a step-rate pause window), and replaying them against the old machine never completes.

Nothing changes from `executing` (a bare standstill without a prior terminal state still concludes nothing), and nothing changes when frames carry `execute` blocks — including wbr!2262 backends, where the block persists. So no previously-completing path behaves differently; the only delta is hang → complete.

PR #488, opened 2026-08-20
feat(NDX-844): Add CommandRoutine execution to the Python SDK

biering:
Entire-Checkpoint: ad06fedab9a5


nova/command_routines/routine.py
```diff
@@ -0,0 +1,97 @@
...
+    Returns a new list; motion commands that already define ``motion_settings`` and
+    non-motion commands are passed through unchanged.
+
+    Examples:
+    >>> from nova.command_routines import move_linear
+    >>> group = with_settings(
+    ...     api.models.MotionSettings(),
+    ...     [move_linear("a")],
+    ... )
+    >>> group[0].motion_settings is not None
```
functionistic: Why is it called `group`?

functionistic (approved):

nova/command_routines/routine.py
biering: fixed

PR #493, opened 2026-08-24
feat: NDX-846 update dataset pose to api

Maddi97:


nova/datasets.py
Maddi97, line 1: Question: I would rather move this under 'cell'

nova/types/__init__.py
```diff
@@ -23,9 +23,8 @@
 __all__ = [
     "Vector3d",
     "Pose",
-    "ConfiguredPose",
-    "DatasetPose",
     "CollisionScene",
+    "Dataset",
```
Maddi97, line 27: I'm only exposing "Dataset"-Model here, not the request config objects (e.g. LoadRemoteDatasetRequest).

My understanding is that this __init__ file is meant to expose the public API to consumers, whereas the request objects are more internal helper types. 

Maybe they don't belong here at all or I should just export them as well

nova/types/dataset.py
Maddi97, line 1: Named this dataset (singular) as this equals pose.py naming. But the module that implements all the wrapper and helper functions is called datasets (plural)

Maddi97:
@claude review

nova/datasets.py
```diff
@@ -0,0 +1,197 @@
...
+    Args:
+        nova: A connected NOVA instance.
+        dataset_id: Restrict the result to all revisions of this dataset.
+        latest_only: When True, return only the latest revision of each dataset.
+    """
+    assert nova.is_connected(), "NOVA instance needs to be connected, in order to list datasets."
+
+    try:
+        return await nova.api.datasets_api.get_datasets(
+            cell=nova.cell().id,
```
Maddi97, line 47: Do we rather want to allow reading from any cell?

Maddi97: Maybe the solution here is to pass an optional cell parameter that defaults to 'CELL_NAME'

Maddi97:
@claude review

nova/program/utils.py
```diff
@@ -47,3 +49,27 @@ def ensure_trailing_newline(s: str):
...
+            if frame.f_code is Program.__call__.__code__:
+                program = frame.f_locals.get("self")
+                if isinstance(program, Program):
+                    code = getattr(inspect.unwrap(program._wrapped), "__code__", None)
+                    program_dir = (
+                        Path(code.co_filename).resolve().parent if code is not None else None
+                    )
+            frame = frame.f_back
+        # TODO should I rather throw an error instead return None
+        return program_dir
```
Maddi97: Rather throw error or return None

nova/datasets/datasets.py
```diff
@@ -0,0 +1,193 @@
...
+        return await nova.api.datasets_api.get_datasets(
+            cell=nova.cell().id,
+            dataset=str(dataset_id) if dataset_id is not None else None,
+            latest_only=latest_only,
+        )
+    except api.ApiException as exc:
+        raise _dataset_error(exc) from exc
+
+
+# TODO: remove maybe or dont put in the api Model rather construct it here
```
Maddi97: Todo & then remove

nova/datasets.py
Maddi97: Why is datasets even under cell

examples/datasets.py
```diff
@@ -0,0 +1,136 @@
...
+from nova.types import MotionSettings
+
+
+async def _move_through_dataset_poses(ctx: nova.ProgramContext, count: int):
+    """Load 'pick' and 'place' poses from the dataset attached to the program and move to them."""
+    assert ctx.dataset is not None and ctx.dataset.poses, (
+        "This program requires a dataset with poses to be loaded."
+    )
+
+    cell = ctx.cell
```
biering: you can remove that line and call it via ctx.cell.controller(... below

examples/datasets.py
```diff
@@ -0,0 +1,136 @@
...
+    motion_group = controller[0]
+    home_joints = await motion_group.joints()
+    tcp_names = await motion_group.tcp_names()
+    tcp = tcp_names[0]
+    pick_pose = ctx.dataset.poses["pick"]
+
+    # Translating a pose between world and local frame
+    pick_pose_fixture = (
+        await ds.transform_to_frame(
+            ctx.nova, [pick_pose.pose], frame="fixture", dataset=ctx.dataset.dataset
```
biering: why do I need to pass ctx.dataset.dataset and not just ctx.dataset?

nova/datasets/datasets.py
```diff
@@ -0,0 +1,193 @@
...
+    except FileNotFoundError as exc:
+        raise DatasetNotFoundError(str(exc)) from exc
+    except (OSError, ValidationError) as exc:
+        raise DatasetError(str(exc)) from exc
+
+    return from_api_model(response)
+
+
+async def list_all(
+    nova: Nova, dataset_id: api.models.DatasetId | None = None, latest_only: bool | None = None
```
biering: since this is anyway a convenience function you could think of passing a str as dataset_id

biering (approved):

nova/datasets/datasets.py
Maddi97: As discussed I deleted those convinience functions anyway

PR #496, opened 2026-08-27
refactor(cell): make MultiTrajectoryCursor a cntrl-style driven cursor

adozzer:
Refactor of the synchronized cursor from PR #489, on top of `feat/trajectory-executor`. No behavior change — a structural rework so `MultiTrajectoryCursor` becomes a true analog of the single `TrajectoryCursor`.

## Why

Two things set the multi cursor apart from the single one:
1. a `TaskGroup` was passed **into** its constructor, and
2. the barrier and the state broadcasters were wired from `attach()`.

The single cursor needs neither: it exposes `cntrl`, which `execute_trajectory` drives, and inside `cntrl` it opens its *own* TaskGroup; `forward()`/`pause()` just feed an intent to that running loop.

## What

- **`control()` — the ensemble's `cntrl` analog.** `attach()` drives it (`task_group.create_task(cursor.control())`) the way `execute_trajectory` drives a single cursor's `cntrl`. Inside `control()` the cursor opens its own `TaskGroup`, spawns the per-group state fan-out + end detachers, and serves an intent mailbox. The barrier runs as a child task there, so pause/supersede cancel it. **No TaskGroup is handed in from outside.**
- **`forward()`/`pause()` feed a mailbox** (`_pending_intent` + `_intent_event`) and return a future — the single cursor's `_set_intent` contract: latest-intent-wins, and `pause()` voids a pending-but-unrun forward.
- **`request_stop()`/`_stop_event`** mirror the single cursor's `detach()`/`_stop_event`: a late `forward()` after teardown fails fast.
- **Broadcasters are created and owned inside the cursor**, not wired by `attach()`. `attach()` only spawns the `executeTrajectory` sockets + session monitors and drives `control()`.

`attach()` shrinks accordingly; the public contract (`forward`/`forward_to`/`pause`/`current_location`/`stream_state`) is unchanged.

## Verification

- Unit: full `tests/cell` + `tests/nova/cell` suites green; the session tests drive everything through the public `attach`/`execute` API, so they pin the reworked internals. `ruff`/`ty` clean (only the pre-existing unrelated `rerun` diagnostic).
- Live: ran `tests/nova/cell/test_trajectory_executor_integration.py` against a virtual KUKA on **SM 26.6.1**. Barrier one-tick start and `forward_to`+step-on pass. The one flake (`pause_then_forward`) is a **controller-side** issue — on this SM a pause of a slow ramp keeps the trajectory `TrajectoryRunning` and never emits `TrajectoryPausedByUser`, so the pause future has no terminal state to complete on. It reproduces **identically on the pre-refactor base**, confirming this refactor introduces no behavior change.

### DIFF
--- a/nova/cell/multi_trajectory_cursor.py
+++ b/nova/cell/multi_trajectory_cursor.py
@@ -9,7 +9,11 @@ from typing import Generic, Protocol, TypeVar
 
 from nova import api
 from nova.actions.io import WriteAction
-from nova.cell.movement_controller.trajectory_cursor import OperationResult, TrajectoryCursor
+from nova.cell.movement_controller.trajectory_cursor import (
+    MovementOption,
+    OperationResult,
+    TrajectoryCursor,
+)
 from nova.core.gateway import ApiGateway
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


adozzer:
Folding the cursor refactor directly into feat/trajectory-executor (PR #489) as a single commit, per request — not merging this PR.

PR #497, opened 2026-08-27
feat(cell): multi-motion-group planning and execution (MultiMotionGroup)

adozzer:
## What

`MultiMotionGroup` — the single primitive for several motion groups, mirroring `MotionGroup` for one: `plan` / `execute` / `plan_and_execute` and one builder. Built on the synchronized `TrajectoryExecutor` from #489 (now in `main`), which this PR also extends.

- **`MultiMotionGroupPlanner`** — the multi-group counterpart of `MotionGroup.plan`. Batches an ensemble action list (`multi_collision_free` / `wait` / `io_write`) into one synchronized `MultiJointTrajectory` via the multi-motion-group RRT endpoint (`search_collision_free_multi_motion_group`), returning a *pure* trajectory (writes carry no joint samples).
- **`MultiMotionGroup`** — the user-facing primitive. Wraps a planner and an executor over the same motion groups (pure DI; the constructor validates both cover the same groups). **One builder** — `MultiMotionGroup.builder(...).sync_on_io(...).build()` — states the sync barrier and wires both halves. A pre-planned trajectory runs through `execute` alone (planner idle), mirroring `MotionGroup.execute`; an all-non-motion list (only writes/waits) is applied directly without planning, the same short-circuit `MotionGroup.plan_and_execute` has.
- **`TrajectoryExecutor`** (extended) — enforces the single-cell invariant itself; carries location-anchored IO from the same `actions` list, routing each write to the owning motion group by its controller and anchoring it on the shared `locations`. The standalone `TrajectoryExecutorBuilder` is gone — assembly moved to `MultiMotionGroup.builder`.
- **`MultiTrajectoryCursor`** — takes the ensemble actions and exposes the single cursor's action stepping (`current_action` / `current_action_index` / `forward_to_next_action`) over the shared parameterization; validates the trajectory's end location matches the motion count, like the single cursor.
- Shared `located_writes` / `write_to_set_io` helpers extracted into `nova.actions.container`.
- Examples: `multi_motion_group.py` (run a recorded synchronized trajectory via `execute`) and `multi_motion_group_planning.py` (plan-and-execute a live collision-free ensemble from fixed joint configs, IO fired on the seam).

## Notes

- Built against api-client 26.7 (plain `list`/`dict`/`float` model shapes; `SetIO.io` unwrapped).
- The planning example is a commented `nova-run-examples` matrix entry — the multi-RRT endpoint is flaky on the CI virtual instance.
- Verified live on a real cluster: the multi-RRT endpoint returns unit-span-per-motion `locations` (so the end-location check holds), and a writes-only `plan` is rejected server-side (`FeedbackCommandsMissing`).

## Test

- New unit suites: `test_multi_motion_group_planner`, `test_multi_motion_group`, `test_located_writes`; plus IO-overlay, action-stepping, direct-non-motion and session coverage in the executor suites, and the live-cell integration test.
- `366 passed` across `tests/actions tests/cell tests/nova/cell`; `ruff format` / `ruff check` / `ty check` clean.

### DIFF
--- a/.github/workflows/nova-run-examples.yaml
+++ b/.github/workflows/nova-run-examples.yaml
@@ -133,6 +133,10 @@ jobs:
             run_cmd: "python examples/multi_motion_group.py"
           - id: bus_io_sync_two_controllers
             run_cmd: "python examples/bus_io_sync_two_controllers.py"
+          # collision-free (RRT) examples are not run in CI: the endpoint is
+          # flaky/unsupported on the virtual instance (see the entry below).
+          # - id: multi_motion_group_planning
+          #   run_cmd: "python examples/multi_motion_group_planning.py"
           # - id: multi_step_movement_with_collision_free
           #   run_cmd: "python examples/multi_step_movement_with_collision_free.py"
           - id: welding
--- a/examples/bus_io_sync_two_controllers.py
+++ b/examples/bus_io_sync_two_controllers.py
@@ -15,7 +15,7 @@ from datetime import datetime
 
 import nova
 from nova import api, run_program
-from nova.cell import GroupArgs, TrajectoryExecutor, virtual_controller
+from nova.cell import GroupArgs, MultiMotionGroup, virtual_controller
 
 CONTROLLERS = ("kuka-a", "kuka-b")
 SYNC_IO = "sync-bus"
@@ -133,8 +133,8 @@ async def bus_io_sync_two_controllers(ctx: nova.ProgramContext):
     await declare_bus_trigger(ctx, gateway, cell.cell_id)
 
     trajectory = await ramp_from_here(groups)
-    executor = (
-        TrajectoryExecutor.builder(groups)
+    ensemble = (
+        MultiMotionGroup.builder(groups)
         .sync_on_io(SYNC_IO, origin=api.models.IOOrigin.BUS_IO)
         .build()
     )
@@ -145,7 +145,7 @@ async def bus_io_sync_two_controllers(ctx: nova.ProgramContext):
     ]
     try:
         tcps = {name: (await group.tcp_names())[0] for name, group in groups.items()}
-        await executor.execute(
+        await ensemble.execute(
             trajectory, groups={name: GroupArgs(tcp=tcp) for name, tcp in tcps.items()}
         )
     finally:
--- a/examples/multi_motion_group.py
+++ b/examples/multi_motion_group.py
@@ -1,9 +1,18 @@
+"""Run a pre-recorded synchronized trajectory with MultiMotionGroup.
+
+``MultiMotionGroup`` is the single primitive for several motion groups — like
+``MotionGroup`` for one. Here it only executes: a trajectory recorded for a robot
+and a positioner is run through the IO barrier with ``MultiMotionGroup.execute``,
+no planning involved. See ``multi_motion_group_planning.py`` for the
+``plan_and_execute`` counterpart.
+"""
+
 from pathlib import Path
 
 import nova
 from nova import api, run_program
 from nova.actions import jnt
-from nova.cell import GroupArgs, TrajectoryExecutor, virtual_controller
+from nova.cell import GroupArgs, MultiMotionGroup, virtual_controller
 
 CONTROLLER = "kuka"
 ROBOT, POSITIONER = f"0@{CONTROLLER}", f"1@{CONTROLLER}"
@@ -53,10 +62,10 @@ async def multi_motion_group_trajectory(ctx: nova.ProgramContext):
         # A trajectory can only be executed from its own first sample.
         await group.plan_and_execute([jnt(samples[name][0])], tcp=TCPS[name])
 
-    executor = (
-        TrajectoryExecutor.builder(groups).sync_on_io(SYNC_IO_ID, controller=CONTROLLER).build()
+    ensemble = (
+        MultiMotionGroup.builder(groups).sync_on_io(SYNC_IO_ID, controller=CONTROLLER).build()
     )
-    await executor.execute(
+    await ensemble.execute(
         trajectory, groups={name: GroupArgs(tcp=tcp) for name, tcp in TCPS.items()}
     )
 
--- a/examples/multi_motion_group_planning.py
+++ b/examples/multi_motion_group_planning.py
@@ -1,0 +1,81 @@
+"""Plan and execute one synchronized ensemble with MultiMotionGroup.
+
+``MultiMotionGroup`` is to several motion groups what ``MotionGroup`` is to one:
+a single object with ``plan`` / ``execute`` / ``plan_and_execute``. Here it plans
+a collision-free move for a robot and a positioner at once (the multi-motion-group
+RRT endpoint), fires an output between the two moves, and runs both groups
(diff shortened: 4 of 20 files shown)


biering:
@claude review

functionistic:
### `multi_collision_free` on the single-group `MotionGroup.plan_and_execute`

Ran into this while reading the PR: since `multi_collision_free()` is exported from `nova.actions` next to `collision_free()`, a user can easily hand it to `MotionGroup.plan_and_execute`. It fails safely, but with an error that never mentions `MultiMotionGroup`:

```
1 validation error for CombinedActions
items.0
  Input tag 'multi_collision_free' found using 'type' does not match any of the
  expected tags: 'linear', 'cartesian_ptp', 'circular', 'joint_ptp', 'spline',
  'collision_free', 'Write', 'Wait'  [type=union_tag_invalid]
```

`MultiCollisionFreeMotion` subclasses `Action` directly rather than `Motion` (`nova/actions/motions.py:370`) while reporting `is_motion() -> True`, so every guard on the single-group path either ignores it or waves it through:

| Step | What happens |
|---|---|
| `_supports_direct_non_motion_actions` (`robot_cell.py:560`) | `False` — not a Wait/Write, no short-circuit |
| `_plan` TCP check (`motion_group.py:908`) | Passes. `_requires_tcp` is `(Linear, CartesianPTP, Circular)`, and the pose-target check is an `isinstance` on the single `CollisionFreeMotion` — so even a pose-target ensemble with `tcp=None` slips through |
| `split_actions_into_batches` (`motion_group.py:43`) | Unrecognized, so it is **glued onto the preceding batch**: `[jnt(...), multi_collision_free(...)]` becomes one batch |
| `validate_collision_setups` (`nova/utils/collision_setup.py:18`) | Passes silently — it filters on `isinstance(action, Motion)`, so the action is invisible and its `collision_setups` (plural) is never inspected |
| `batch[0]` dispatch (`motion_group.py:932`) | Neither `CollisionFreeMotion` nor `WaitAction` → falls to the `else` → `_plan_with_collision_check` |
| `CombinedActions(items=tuple(actions))` (`motion_group.py:774`) | 💥 the tag is not in the `Motion \| WriteAction \| WaitAction` union |

`MultiCollisionFreeMotion.to_api_model()` does raise a clear `NotImplementedError`, but it is dead code on this path — pydantic construction fails first.

Two notes:

- **Ordering changes the cost.** After a `collision_free`, batching splits, so batch 1 issues a real RRT request to the server and only batch 2 raises. Planning work is spent and discarded. Nothing executes either way — the failure is strictly pre-execution.
- **No silent misbehavior.** I checked whether it could be quietly dropped instead (the way `to_motion_command` skips `CollisionFreeMotion`) — it cannot; validation fails at container construction, so there is no path where the ensemble motion vanishes and the robot moves somewhere unexpected.

Not a blocker, and entirely your call whether it belongs in this PR. If you want a guardrail, the cheapest one is an explicit reject in `MotionGroup._plan` alongside the existing TCP-requirement checks, pointing at `MultiMotionGroup.plan_and_execute` — that also catches it before the wasted RRT round-trip.

functionistic:
### The `IOSyncDriver` bus-IO poll — can it be push instead?

Separate topic from my last comment, and to be clear up front: **this predates this PR.** The loop came in with #489 (`7a1ea104`) and #497 doesn't touch `_write`. Raising it here because #497 is what puts this machinery behind a public `MultiMotionGroup` API, so it's the point where the behaviour becomes something users depend on.

#### Where it is

One loop in the whole barrier path — the `BUS_IO` branch of `IOSyncDriver._write` (`nova/cell/multi_trajectory_cursor.py:130-136`):

```python
await self._api_client.bus_ios_api.set_bus_io_values(cell=..., io_value=[io_value])
while True:
    values = await self._api_client.bus_ios_api.get_bus_io_values(cell=..., ios=[action.key])
    if values and values[0] == io_value:
        break
    await asyncio.sleep(_IO_POLL_INTERVAL)   # 50 ms
```

The `CONTROLLER` branch immediately below is already push — a single `wait_for_io_event` the server holds open. So this isn't a design choice, it's an **API asymmetry**:

| | `ControllerInputsOutputsApi` | `BusInputsOutputsApi` |
|---|---|---|
| set | `set_output_values` | `set_bus_io_values` |
| read | `list_io_values` | `get_bus_io_values` |
| **wait** | `wait_for_io_event` | — |
| **stream** | `stream_io_values` | — |

#### `clear()` needs the confirmation; `release()` doesn't

`_forward` runs `clear()` → arm all cursors → `_wait_until_waiting_for_io` (already push, via the state stream) → `release()`. The two calls are not equally load-bearing:

- **`clear()`** — correctness. If a group arms while the IO image still reads *released*, it starts immediately and the run desynchronizes. Keep it.
- **`release()`** — not correctness. Every group is already armed by then; the write itself is what starts them, observed or not. The read-back only buys failure detection (a dropped write surfaces in 5 s instead of hanging in the final `gather`) — and it costs up to 50 ms of dead time at exactly the wrong moment, between "all armed" and "moving".

So roughly half the polling can go with no correctness argument against it. A single delayed re-read (one request ~250 ms later) would keep the failure signal without a loop.

#### The push replacement already exists in the SDK

`nova/utils/io.py:95-162` — `wait_for_bus_io()` subscribes to `nova.v2.cells.{cell}.bus-ios.ios` and decodes `list[IOValue]` per message. It already implements the tricky part: subscribe first, *then* read the current value, then consume (`io.py:141-151`) — the NDX-665 fix from #413, which closes exactly the race the poll avoids by brute force.

Two obstacles, both plumbing rather than design:

1. `IOSyncDriver` has no NATS handle — it gets `api_client: ApiGateway` + `cell: str`. The client lives on `Cell` (`cell.py:60`) and `MotionGroup` stores only the cell *name* (`motion_group.py:152`), so `MultiMotionGroupBuilder.build()` — which harvests `barrier_motion_group._api_client` / `._cell` — can't reach it.
2. `wait_for_bus_io` is written against `Nova` (`_resolve_nova`, `nova.api`, `nova.nats`) rather than `(nats, api_client, cell)`. Reuse means lifting its core onto the lower-level triple, which is the right refactor anyway since `Cell` already holds both halves.

Caveats: its own docstring says at-most-once delivery, so the 5 s `asyncio.timeout` stays either way; and since NATS may not be connected, you'd be choosing between hard-failing a bus-IO barrier without it or falling back to the poll.

#### The proper fix is server-side

A `wait_for_bus_io_event` mirroring `wait_for_io_event` collapses the `BUS_IO` branch into the same three lines as the `CONTROLLER` branch — no NATS dependency, no builder plumbing, no subscribe/read race to reason about. The table above is a decent argument for it on its own. That's an API ask, not SDK work.

#### Suggested order

1. **Drop the read-back on `release()`** — free, removes the poll from the latency-critical window.
2. **Ask for `wait_for_bus_io_event`** — the real fix; worth filing now since it unblocks the rest.
3. **NATS for `clear()`** — correct today and reuses existing code, but costs the `Cell`/NATS plumbing through the builder.
4. **Prefer `origin=CONTROLLER` where the topology allows** — that path already polls zero times. Not a fix, but it limits exposure.

None of this needs to happen in #497 — flagging it so the decision is explicit rather than inherited.

functionistic (approved):

biering (approved):
lgtm

PR #510, opened 2026-09-07
fix(RB-4251): J5 transform for KR6 R1820

JulScho:
image
this fixes the offset in the link_5 rotations for a bunch of kukas and a bunch of abbs. it reintroduces a fault in the hc10dtp but this gets fixed in parallel in the rdp: [https://code.wabo.run/robotics/robot-data-provider/-/merge_requests/280](https://code.wabo.run/robotics/robot-data-provider/-/merge_requests/280)

This also adds support for glbs with kinematic chain offset

### DIFF
--- a/nova_rerun_bridge/dh_robot.py
+++ b/nova_rerun_bridge/dh_robot.py
@@ -3,19 +3,70 @@ import numpy as np
 from nova import api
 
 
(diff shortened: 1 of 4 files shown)


stefanwagnerdev (approved):
Will soon be replaced with URDF

PR #515, opened 2026-09-16
feat(NDX-872): expose all API blending options

biering:
## Summary
- Add `MotionSettings.blending`, typed as the API's `BlendingAuto | BlendingPosition` discriminated union, so every blending option the NOVA API offers is reachable from the SDK.
- Deprecate `blending_auto` / `blending_radius`; they keep working and still map onto the same API messages.
- Bump `wandelbots_api_client` from `26.7.0.dev24` to `26.7.0.dev65`.

## Why
Requested by Jonas (PS) for a customer. The API's `BlendingPosition` message carries seven options, but `MotionSettings` only ever surfaced two of them — `position_zone_radius` (via `blending_radius`) and `min_velocity_in_percent` (via `blending_auto`). Anyone needing percentage-based zones, orientation blending, or joint-space blending had to bypass `MotionSettings` and hand-build `api.models.MotionCommand`.

## Benefits
- `position_zone_percentage`, `orientation_zone_radius`, `orientation_zone_percentage`, `joints_zone_radius`, `joints_zone_percentage` and `space` (`JOINT` / `CARTESIAN`) are now settable per action through the normal `settings=` path.
- Because the field *is* the API model rather than an SDK mirror, future spec additions land in the SDK with a client bump instead of a code change.
- No migration needed: existing programs using `blending_radius` / `blending_auto` / `position_zone_radius` behave exactly as before.

```python
settings = MotionSettings(
    blending=api.models.BlendingPosition(
        joints_zone_radius=0.1,
        joints_zone_percentage=20.0,
        space=api.models.BlendingSpace.JOINT,
    )
)
```

## How
`as_blending_setting()` returns the new field when set and otherwise falls back to the legacy construction, so the single `api.models.MotionCommand` call site in `nova/actions/container.py` needed no change. The validator rejects mixing the new field with any legacy one.

Two non-obvious bits:

- Pydantic's `deprecated=True` warns on **every** read, including reads from inside the model's own methods. `_get_blending_radius()` / `_get_blending_auto()` now read via `self.__dict__`, otherwise a plain `MotionSettings()` would emit four `DeprecationWarning`s per construction. This also silences pre-existing noise from `position_zone_radius` / `min_blending_velocity`.
- `wandelscript/builtins/__init__.py` auto-registers a builtin per `MotionSettings` field name *and* separately aliases `position_zone_radius` → `blending(n)`. A field named `blending` collided, so the loop now skips it and the wandelscript `blending(n)` float API is unchanged.

Collision-free motions still do not support blending — `PlanCollisionFreeRequest` has no such field.

## Test plan
- [x] `uv run ruff format`, `uv run ruff check .`, `uv run ty check` — clean
- [x] `PYTHONPATH=. uv run pytest -q -m "not integration"` — 934 passed, 6 skipped
- [x] New tests in `tests/nova/types/test_motion_settings.py`: full seven-option round-trip, `BlendingAuto`, new-vs-legacy conflict matrix, JSON round-trip, and that the complete message reaches `api.models.MotionCommand`
- [x] Verified against the installed `26.7.0.dev65` client that all seven `BlendingPosition` fields exist and that `MotionCommand.blending` is discriminated on `blending_name`
- [ ] `examples/blending.py` not executed against a live NOVA instance

## Notes for reviewers
- The doctests in `nova/actions/motions.py` embed the full `MotionSettings` repr, so 10 of them needed `blending=None` inserted. That is churn, not behaviour.
- `pyproject.toml` also drops Andreas Langenhagen from `authors` — unrelated to blending, included because it was already in the working tree. Say the word and I'll pull it out.
- Deliberately left out: extending the wandelscript `blending()` builtin to the full option set, and segment blending in `examples/merge_trajectories.py` (already uses the API model directly).


functionistic (approved):

wandelscript/builtins/__init__.py
mkhe93, line 1: do we really need this change in wandelscript?

mkhe93 (changes requested):

mkhe93 (approved):

wandelscript/builtins/__init__.py
biering: yes, I wanted to check if anything breaks in the new version of the API client (we have a lot of tests in ws) and Andreas is no maintainer anymore (not at WB).

PR #521, opened 2026-09-18
fix(cursor): ignore the previous stop's re-published terminal state on resume

functionistic:
## Summary
- Make `TrajectoryExecutionMachine` ignore frames that merely repeat the terminal state a `start` leaves (same kind at the same location) until the controller shows any different frame.
- Pin the behaviour with machine-level tests and a cursor-level replay of the observed frame sequence; document the rule in the movement-controller README.

## Why
Since 2026-09-18 every branch's integration CI failed `test_forward_to_intermediate_target_keeps_the_session_open`: after `forward_to(1.0)`, the follow-up `forward()` resolved at ~1.0 while the robot ran on to 2.0. The portal now provisions a controller build that publishes execute state level-based (robotics/wbr!2262): after the stop at 1.0 it re-publishes `END_OF_TRAJECTORY@1.0` every step until it takes up the next start (then `WAIT_FOR_IO → RUNNING → END@2.0`). The cursor sent `start` for the resume, saw those stale END frames first and concluded the new operation at its own start location. The `paused` path already guarded against this (`may_complete_as_paused`); the `ended` path did not. Not caused by any PR — reproduced with the SDK code and api-client of the last green run.

## Benefits
- Stepping on after an intermediate target, and every resume from `ended`/`paused`, reports the real stop again — the executor session and `MultiTrajectoryCursor` stay usable on current controllers.
- No behaviour change on controllers that drop the `execute` block at settle; nothing changes for a first start from `idle`.

## How
`on_start` records the (kind, location) of the terminal frame being left; `process_motion_state` treats frames matching it as `_keep_executing` until a different frame arrives, which lifts the filter. Identity by kind and location rather than a "seen RUNNING" guard is deliberate: a `forward()` issued at the very end answers `WAIT_FOR_IO@2.0` then `END@2.0` with no RUNNING frame in between (probed live), and a running guard would hang there.

## Test plan
- [x] Live on a fresh NOVA instance (172.31.13.138): `PYTHONPATH=. uv run pytest -m integration tests/nova/cell/test_trajectory_executor_integration.py` — 3 passed (fails 1.002 ≥ 1.99 without the fix)
- [x] Live: `tests/nova/cell/motion_group/test_trajectory_cursor_integration.py`, `test_pause_on_io.py`, `test_motion_group_movement.py` — 10 passed
- [x] Live probe of raw `stream_state()` frames around the resume and around a `forward()` issued at the end (WAIT_FOR_IO then END again, no RUNNING)
- [x] `PYTHONPATH=. uv run pytest -m "not integration"` — 1016 passed; 7 new tests (stale END/PAUSED ignored, filter lifts on WAIT_FOR_IO/RUNNING, start at the end, no filter from idle, cursor-level `forward_to(1.0)` then `forward()` replay)
- [x] `uv run ruff format`, `uv run ruff check`, `uv run ty check`


functionistic:
@claude review

PR #529, opened 2026-09-23
feat(datasets)!: resolve dataset frames locally via as_world()

biering:
## Summary
- Resolve dataset frame chains locally and expose `as_world()` on dataset poses and frames; delete `transform_to_frame` / `transform_to_world`.
- Reject a frame-relative `DatasetPose` used as a motion target instead of silently using the unresolved pose.
- Add `Dataset.set_frame()` so a frame measured at runtime (camera) rebinds everything taught on it.

## Why

A dataset pose taught relative to a frame was handed out as if it were a world pose. `cartesian_ptp(ctx.dataset.poses["fixture-slot-a"])` compiled, ran, and sent the robot to the wrong place — no error, no warning. Resolving it properly meant an API round trip per transform, with `[0]` indexing and a `Nova` handle threaded through every call site.

That round trip was never necessary. `DatasetFrame.reference_frame` chains up to `world` and `DatasetPose.frame` names the frame a pose was taught in — the whole tree already ships inside the dataset. `grep reference_frame` across the repo hit only `examples/example_dataset.json`: the SDK had never read it.

## Benefits

- Frame-relative poses can no longer reach a robot unresolved — they either resolve or raise.
- Pose transformation is synchronous and free: no `await`, no network, no `Nova` handle. A 24-slot palletizing grid is plain `@` arithmetic.
- A pallet measured per cycle rebinds one frame and every slot below it follows, which is the actual reason to keep a grid in frame coordinates.

## How

`FrameTree` walks `reference_frame` to world and composes with `@`. Resolution is **lazy**: a dataset with a dangling frame reference still loads, and only the entries on the broken chain raise `FrameResolutionError` (new, carries the frame id and the chain walked). Sibling poses keep working.

Nothing is cached. An earlier revision memoized composed transforms and returned a stale pose after `set_frame()` replaced a frame — silently, which is exactly the failure mode this PR exists to remove. Chains are a few links deep, so the cache bought nothing. For the same reason `FrameTree` reads frames off the `Dataset` rather than holding its dict, so replacing `dataset.frames` wholesale cannot strand it.

`.pose` deliberately keeps its wire meaning rather than becoming the world pose. That keeps the SDK types as thin subclasses of the generated models and leaves `model_dump()` round-tripping the wire shape.

**Known limit:** `cartesian_ptp(p.pose)` on a frame-relative pose is still silently wrong. Once unwrapped to a plain `Pose` there is no frame left to check — only passing the object, `cartesian_ptp(p)`, can raise. Both examples always use `as_world()`.

## Breaking change / migration

`nova.datasets.transform_to_frame` and `transform_to_world` shipped in v6.0.0 and are removed.

```python
# before
world = (await ds.transform_to_world(nova, dataset_id, [p.pose], frame="fixture"))[0]
local = (await ds.transform_to_frame(nova, dataset_id, [world], frame="fixture"))[0]

# after
world = dataset.poses["fixture-slot-a"].as_world()
local = ~dataset.frames["fixture"].as_world() @ world
```

Passing a frame-relative `DatasetPose` to a motion action now raises `ValueError` naming the frame. Previously it used the taught pose. Any caller that relied on the old behaviour was already producing a wrong target.

## Test plan

- [x] `PYTHONPATH=. uv run -p 3.12 --all-extras pytest -q -m "not integration"` — 1039 passed, 6 skipped (1020 on `main`)
- [x] `uv run -p 3.12 ruff format . && ruff check --select I --fix . && ruff check . && ty check nova` — clean
- [x] `PYTHONPATH=. uv run -p 3.12 python examples/pose_transformations.py` — every asserted number matches
- [x] Both examples run against `examples/example_dataset.json` and `palletizing_dataset.json`: 3-level chain (`table` → `fixture` → `tool-frame`) resolves, `~frame @ world` round-trips back to `.pose`, and the measured pallet shifts all 24 slots
- [x] Error paths: dangling `frame`, dangling `reference_frame`, `a -> b -> a`, self-reference `a -> a`, detached entry after `model_dump()`/`model_validate()`
- [ ] **`tests/nova/datasets/test_frames_integration.py` has not been run** — it needs a live cell

## Notes for reviewers

Focus on `nova/datasets/frames.py`. It replaces a server-side computation with local math, and the integration test above is the only thing that would catch a convention mismatch between the two. Please run it against a cell before merging.

`examples/pose_transformations.py` is a standalone script (no robot, no NOVA connection) that asserts its own expected numbers, so it doubles as a spec for `@` and `~`. It and `examples/palletizing.py` are wired into the examples CI matrix.


nova/datasets/types.py
```diff
@@ -1,10 +1,84 @@
...
+    """A dataset pose whose `pose` is an SDK `Pose` instead of the API wire model.
+
+    `pose` is the pose as taught, relative to `frame`. Use `as_world()` to resolve it
+    into world coordinates.
+    """
+
+    # `from_attributes` lets an `api.models.DatasetPose` be validated into this type as-is.
+    model_config = ConfigDict(from_attributes=True)
+
+    pose: SdkPose
```
Maddi97: Can this be type Pose (from nova.types) or what advantages bring us this new type?

biering: good point, I removed it

nova/datasets/types.py
```diff
@@ -1,10 +1,75 @@
...
+    transform from this frame into world coordinates.
+    """
+
+    model_config = ConfigDict(from_attributes=True)
+
+    pose: Pose
+
+    _tree: FrameTree | None = PrivateAttr(default=None)
+
+    def as_world(self) -> Pose:
```
Maddi97: Idea: Could be cached in the future

